PR #2, opened 2017-04-07
Gb long constraints support

ewelinagr:
Fix GBD-4 - an issue with URL length limit
Adding support for both POST and GET for calls that can reach this limit


ewelinagr:
It's a copy of this PR: https://github.com/thehyve/transmart-core-fork/pull/15

PR #6, opened 2017-04-10
Add transmart-server as a rest-only version of tranSMART

ewelinagr:
- Add transmart-server as a main grails application, which exposes the REST API and handles the OAuth flow
- Change transmartApp into a grails plugin
- Convert shared authentication code in transmart-app and transmart-server into seperate module: transmart-oauth


transmart-oauth/grails-app/domain/org/transmart/oauth/Client.groovy
```diff
@@ -0,0 +1,68 @@
...
+        autoApproveScopes nullable: true
+
+        redirectUris nullable: true
+        additionalInformation nullable: true
+
+        //oAuthIDs nullable: true
+    }
+
+    def beforeInsert() {
+        encodeClientSecret()
```
JanKanis: There was a check here (removed in commit 040336b87e55665406ac70969a3910c8c2b63a3f) to remind the user to remove a workaround if a bug in grails was fixed. AFAIK that bug is still present in grails 3.2.3.

JanKanis: Actually not actively removed, but it looks like it wasn't copied from transmartApp with all the rebasing going on.

transmart-server/grails-app/domain/org/transmart/oauth2/OAuthGoogle.groovy
```diff
@@ -0,0 +1,42 @@
...
+
+import org.transmart.oauth.Client
+
+/**
+ * Simple domain class that records the identities of users authenticating via
+ * an OAuth provider. Each identity consists of the OAuth account name and the
+ * * name of the provider. It also has a reference to the corresponding Spring Security
+ * user account, although only long IDs are supported at the moment.
+ */
+class OAuthGoogle implements Serializable {
```
JanKanis: Should this class still be in here?

transmart-server/build.gradle
```diff
@@ -0,0 +1,64 @@
...
+    profile "org.grails.profiles:web"
+    runtime "com.bertramlabs.plugins:asset-pipeline-grails:2.11.6"
+    runtime "com.h2database:h2"
+    testCompile "org.grails:grails-plugin-testing"
+    testCompile "org.grails.plugins:geb"
+    testRuntime "org.seleniumhq.selenium:selenium-htmlunit-driver:2.47.1"
+    testRuntime "net.sourceforge.htmlunit:htmlunit:2.18"
+    // end of default Grails 3.2 deps
+
+    compile 'org.grails.plugins:spring-security-core:3.1.1'
```
JanKanis: Why 3.1.1 instead of 3.2.3?

ewelinagr: 3.1.1 is the latest I found in their repo:
https://github.com/grails-plugins/grails-spring-security-core
Are you sure there is 3.2.3 version already?

transmart-oauth/grails-app/services/org/transmart/oauth/authentication/BruteForceLoginLockService.groovy
```diff
@@ -40,7 +40,7 @@ class BruteForceLoginLockService {
      */
     def failLogin(String login) {
         def numberOfAttempts = failedAttempts.get(login)
-        log.debug "fail login ${login} previous number for failedAttempts $numberOfAttempts"
+        BruteForceLoginLockService.log.debug "fail login ${login} previous number for failedAttempts $numberOfAttempts"
```
JanKanis: why this change?

transmart-oauth/src/main/groovy/org/transmart/oauth/LdapAuthUserDetailsMapper.groovy
```diff
@@ -12,13 +13,11 @@ import org.springframework.util.Assert
...
+import org.transmart.oauth.authentication.AuthUserDetailsService
 
-/**
- * User: Florian Guitton
- * Date: 15/10/13
- * Time: 11:23
- */
 public class LdapAuthUserDetailsMapper implements UserDetailsContextMapper {
+    @Autowired
+    AuthUserDetailsService userDetailsService
```
JanKanis: This doesn't seem to be used anywhere

transmartApp/grails-app/services/org/transmartproject/security/OAuth2SyncService.groovy
```diff
@@ -62,7 +61,7 @@ class OAuth2SyncService {
                     /* Workaround to bug https://github.com/grails/grails-core/issues/10451 due to which event handlers such as beforeInsert()/beforeUpdate() are
                      * not called on domain objects of non-default datastores. Remove this workaround if the bug has
                      * been solved in a grails update, and remove the corresponding exception code in
-                     * org.transmart.oauth2.Client.
+                     * org.transmart.oauth.Client.
                      */
                     client.encodeClientSecret()
                 }
```
JanKanis: I think this class also should move to transmart-oauth

transmartApp/src/main/groovy/org/transmart/FederatedUserDetailsService.groovy
```diff
@@ -1,6 +1,5 @@
 package org.transmart
```
JanKanis: This class (and SAML authentication in general) should probably also be moved to transmart-oauth. But it would be difficult to test if it still works...

transmart-server/build.gradle
JanKanis: Ah, I would expect the versions to match the grails version. Especially since the other two spring security modules below are at 3.2.3

JanKanis:
If I start transmart-server and go to /, I get an error instead of being redirected to a useful page

transmart-server/grails-app/controllers/org/transmart/server/UrlMappings.groovy
```diff
@@ -0,0 +1,18 @@
...
+class UrlMappings {
+
+    static mappings = {
+        "/$controller/$action?/$id?(.$format)?"{
+            constraints {
+                // apply constraints here
+            }
+        }
+
+        "/"(controller: 'auth', action: 'index')
```
JanKanis: Doesn't this require an AuthController to exist? Navigating to `/` gives an error.

transmart-oauth/grails-app/services/org/transmart/oauth/authentication/AuditLogService.groovy
```diff
@@ -0,0 +1,88 @@
...
+
+import javax.servlet.http.HttpServletRequest
+
+/**
+ * Logs to level TRACE in a specific `metric' format.
+ * Configuration of output done in (out-of-tree) Config.groovy.
+ */
+@CompileStatic
+@Slf4j
+class AuditLogService {
```
JanKanis: This class now exists in transmart-oauth and transmartApp. Also this is not authentication related, so transmart-oauth is not the right place for this to live. Maybe the right place for this is transmart-server.

transmart-server/grails-app/conf/logback.groovy
```diff
@@ -0,0 +1,39 @@
+import grails.util.BuildSettings
```
JanKanis: This is the auto generated logback conf. Probably better to copy the one from transmartApp, as that has some extra stuff in it. (I presume grails will take the logback.groovy from the main application module)

transmart-oauth/grails-app/conf/application.yml
```diff
@@ -0,0 +1,176 @@
...
+            password: biomart_user
+            dbCreate: none
+    test:
+        dataSource:
+            dbCreate: update
+            url: jdbc:h2:mem:testDb;MVCC=TRUE;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE
+    production:
+        dataSource:
+            dbCreate: none
+            url: jdbc:h2:./prodDb;MVCC=TRUE;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE
```
JanKanis: This doesn't look correct for a production setup...

JanKanis: @PiotrZakrzewski where does this come from?

PiotrZakrzewski: @JanKanis It must be Grails generated boilerplate

transmartApp/src/main/groovy/transmartapp/TransmartAppGrailsPlugin.groovy
```diff
@@ -0,0 +1,171 @@
...
+                        scheme = grailsApplication.config.org.transmart.security.saml.lb.scheme
+                        serverName = grailsApplication.config.org.transmart.security.saml.lb.serverName
+                        serverPort = grailsApplication.config.org.transmart.security.saml.lb.serverPort
+                        includeServerPortInRequestURL = grailsApplication.config.org.transmart.security.saml.lb.includeServerPortInRequestURL
+                        contextPath = grailsApplication.config.org.transmart.security.saml.lb.contextPath
+                    }
+                } else {
+                    contextProvider(org.springframework.security.saml.context.SAMLContextProviderImpl)
+                }
+            }
```
JanKanis: needs to be moved if we move SAML to tm-server

transmartApp/src/main/groovy/transmartapp/TransmartAppGrailsPlugin.groovy
```diff
@@ -0,0 +1,171 @@
...
+
+            redirectStrategy(DefaultRedirectStrategy)
+
+            accessDeniedHandler(AccessDeniedHandlerImpl) {
+                errorPage = '/login'
+            }
+            failureHandler(SimpleUrlAuthenticationFailureHandler) {
+                defaultFailureUrl = '/login'
+            }
```
JanKanis: Maybe this block belongs in tm-oauth?

transmart-oauth/src/main/groovy/org/transmart/oauth/TransmartOauthGrailsPlugin.groovy
```diff
@@ -0,0 +1,174 @@
...
+
+    // Online location of the plugin's browseable source code.
+//    def scm = [ url: "http://svn.codehaus.org/grails-plugins/" ]
+
+    Closure doWithSpring() {
+        { ->
+            xmlns context: "http://www.springframework.org/schema/context"
+            xmlns aop: "http://www.springframework.org/schema/aop"
+
+            sessionRegistry(SessionRegistryImpl)
```
JanKanis: TBD: Do we need to configure a concurrent sessions limit like transmartApp does?

transmart-oauth/src/main/groovy/org/transmart/oauth/TransmartOauthGrailsPlugin.groovy
```diff
@@ -0,0 +1,174 @@
...
+                        userDetailsContextMapper = ref('ldapUserDetailsMapper')
+                    }
+                }
+            }
+
+            bruteForceLoginLockService = ref('bruteForceLoginLockService')
+            bruteForceLoginLockService(BruteForceLoginLockService) {
+                allowedNumberOfAttempts = grailsApplication.config.bruteForceLoginLock.allowedNumberOfAttempts
+                lockTimeInMinutes = grailsApplication.config.bruteForceLoginLock.lockTimeInMinutes
+            }
```
JanKanis: To verify: Do these settings also apply to transmartApp sessions? (as that configures its own sessionRegistry)

transmart-oauth/grails-app/init/org/transmart/oauth/Application.groovy
```diff
@@ -0,0 +1,16 @@
...
+
+import grails.boot.*
+import grails.boot.config.GrailsAutoConfiguration
+import grails.plugins.metadata.*
+import groovy.util.logging.Slf4j
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration
+import org.springframework.boot.autoconfigure.security.SecurityFilterAutoConfiguration
+
+@Slf4j
+@EnableAutoConfiguration(exclude = [SecurityFilterAutoConfiguration])
```
JanKanis: The other plugins have `@PluginSource` here instead of these annotations. Is this correct?

JanKanis:
There's also stuff in the root build.gradle that is specific to plugins or to stand alone apps, that also needs to be updated. 

I would like someone with more gradle knowledge to review the build.gradle changes (and tell me what everything means that I can't find in the docs.)

transmartApp/grails-app/init/org/transmartproject/app/BootStrap.groovy
```diff
@@ -50,11 +50,6 @@ class BootStrap {
 
         // force marshaller registrar initialization
         grailsApplication.mainContext.getBean 'marshallerRegistrarService'
-
-        if ('clientCredentialsAuthenticationProvider' in
-                grailsApplication.config.grails.plugin.springsecurity.providerNames) {
-            OAuth2SyncService.syncOAuth2Clients()
-        }
     }
```
JanKanis: The blocks about SAML probably also should be in transmart-server. I don't know what the purpose of the forceEagerSessionCreation and concurrent sessions stuff is, so I'm not sure if it belongs here or in -server.

There's also RScript related config that should ideally live in the R script module, but we don't need to bother with that now.

I would think the configFine check should live in tm-server (but I'm not certain). Currently it is there too, duplicated. Todo: deduplicate it. And so is the fixupConfig stuff, that probably shouldn't be in there. Does grails run the BootStrap.init's from all modules or just from the main application?

And tm-server Bootstrap.groovy and tm-oauth Bootstrap.groovy are exact copies. That also needs deduplication. I think tm-oauth doesn't need most of what's in there.

ewelinagr:
@JanKanis I also wasn't sure about ChildProcessAppender
https://github.com/thehyve/transmart-core/blob/rest-only/transmartApp/src/main/groovy/org/transmart/logging/ChildProcessAppender.groovy
There is some configuration for this in transmartApp/logback.groovy 
Do you think it should be moved to tm-server?

JanKanis:
@ewelinagr ChildProcessAppender can be moved, together with JsonLayout in org.transmart.logging and their associated unit tests. (The unit tests in transmartApp should work, unfortunately the integration tests don't.) There is also a test for OAuth2SyncService.

JanKanis:
Having the AuditLogService in tm-server will create a dependency of transmartApp and other modules on tm-server but I don't think that's a big problem. Creating a transmart-logging module would be a bit overkill.

transmart-oauth/grails-app/init/org/transmart/oauth/Application.groovy
ewelinagr: you're right, it should be @PluginSource here. @EnableAutoConfiguration should be added only to transmart-server/Application.groovy (as a workaround to some grails-spring-security-core plugin issue), I tested this and it is not breaking after a change in transmart-oauth/Application.groovy

transmartApp/grails-app/controllers/org/transmartproject/interceptors/ConceptsInterceptor.groovy
```diff
@@ -1,12 +1,11 @@
 package org.transmartproject.interceptors
 
-import org.transmart.audit.AuditLogService
 import org.transmartproject.core.users.User
 
 
 class ConceptsInterceptor {
 
-    AuditLogService auditLogService
+    def auditLogService
```
JanKanis: This makes it do by-name bean insertion, by type is more robust. Use an @Autowired if needed. You can also update the studyIdService bean in the next line while you're at it.

transmartApp/grails-app/controllers/org/transmartproject/interceptors/DataExportInterceptor.groovy
```diff
@@ -11,7 +10,7 @@ import org.transmartproject.core.users.User
 class DataExportInterceptor implements Interceptor {
 
     AccessLogEntryResource accessLogService
-    AuditLogService auditLogService
+    def auditLogService
```
JanKanis: same here

transmart-oauth/grails-app/services/org/transmart/oauth/authentication/BruteForceLoginLockService.groovy
```diff
@@ -40,7 +40,7 @@ class BruteForceLoginLockService {
      */
     def failLogin(String login) {
         def numberOfAttempts = failedAttempts.get(login)
-        BruteForceLoginLockService.log.debug "fail login ${login} previous number for failedAttempts $numberOfAttempts"
+        log.debug "fail login ${login} previous number for failedAttempts $numberOfAttempts"
         failedAttempts.put(login, numberOfAttempts + 1)
     }
```
JanKanis: Don't forget the next instance in loginSuccess below vv

transmartApp/grails-app/controllers/org/transmartproject/interceptors/ConceptsInterceptor.groovy
ewelinagr: Adding by-type spring bean insertion will force adding dependency on transmart-server, so we will have a circular dependency there (between tm-app and tm-server). Maybe it's better to move this bean definition somewhere else?

transmart-server/grails-app/conf/logback.groovy
ewelinagr: ok, I changed it as you suggested (although github not marking this as outdated)

transmartApp/grails-app/controllers/org/transmartproject/interceptors/ConceptsInterceptor.groovy
JanKanis: The AuditLogService is used to track events in an audit log. There are several interceptors that call into it, though now that I look at it they are all in transmartApp except for one usage in transmart-gwas-plugin. However one is the OauthInterceptor (that also needs to be moved I think).

Using a by type or by name bean is a dependency either way. The difference is that in one case it needs to be declared. Without further changes the interceptors that use this bean by-name will fail at runtime if it is not set.

The way the code is organised now this is a hard problem. Both transmartApp and transmart-oauth will need to use this service. I guess the architecturally correct solution is to create a new module for this, but I don't like the overhead of that for just a couple of classes.

Maybe the best solution is to create an AuditLogger interface for it in core-api and use that in the modules that use it with @Autowired(required=false), and make sure the code does not throw exceptions if it is not available. In that way the service can live in tm-server without a problem.

transmart-server/build.gradle
```diff
@@ -0,0 +1,67 @@
+version "17.1"
+group "transmart.server"
```
gijskant: version, group and most of the plugins should be set by the root `build.gradle`, not in the subproject build file.

transmart-core-api/src/main/groovy/org/transmartproject/core/audit/AuditLogger.groovy
```diff
@@ -0,0 +1,10 @@
+package org.transmartproject.core.audit
+
+import javax.servlet.http.HttpServletRequest
+
+interface AuditLogger {
+
+    boolean getEnabled()
+
+    def report(Map<String, Object> params, String event, HttpServletRequest request)
+}
```
PiotrZakrzewski: Small stylistic feedback: turn on automatic empty line adding in IntelliJ https://www.jetbrains.com/help/idea/code-style-java.html
Why it is important: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline#729795

transmart-oauth/grails-app/conf/application.yml
```diff
@@ -0,0 +1,176 @@
...
+                timeBetweenEvictionRunsMillis: 5000
+                minEvictableIdleTimeMillis: 60000
+                validationQuery: SELECT 1
+                validationQueryTimeout: 3
+                validationInterval: 15000
+                testOnBorrow: true
+                testWhileIdle: true
+                testOnReturn: false
+                jdbcInterceptors: ConnectionState
+                defaultTransactionIsolation: 2 # TRANSACTION_READ_COMMITTED
```
PiotrZakrzewski: Same as above

transmart-oauth/grails-app/controllers/org/transmart/oauth/interceptors/OauthInterceptor.groovy
```diff
@@ -1,12 +1,18 @@
...
     //matches OauthController and all actions of controller by naming convention
-    def auditLogService
+
+    @Autowired(required = false)
+    AuditLogger auditLogService
+    @Autowired
     User currentUserBean
+
     boolean before() {
         auditLogService.report("OAuth authentication", request,
```
PiotrZakrzewski: I might be wrong - but will not cause null pointer exception here if auditLogService fails to get wired (like it is allowed to since it is not required)?

transmart-oauth/grails-app/init/org/transmart/oauth/BootStrap.groovy
```diff
@@ -0,0 +1,177 @@
...
+                    "with a breakpoing on the catch block in ConfigurationHelper::mergeInLocations().\n" +
+                    "Alternatively, you can change the console logging settings by editing " +
+                    "\$GRAILS_HOME/scripts/log4j.properties, adding abecause it does not reside in the file system proper appender and log " +
+                    "org.codehaus.groovy.grails.commons.cfg.ConfigurationHelper at level WARN")
+            throw new GrailsConfigurationException("Configuration magic setting not found")
+        }
+
+        fixupConfig()
+
+//        // force marshaller registrar initialization
```
PiotrZakrzewski: We could as well remove this now

transmart-oauth/grails-app/services/org/transmart/oauth/OAuth2SyncService.groovy
```diff
@@ -59,10 +58,10 @@ class OAuth2SyncService {
                 dirty = true
 
                 if (prop == 'clientSecret') {
-                    /* Workaround to bug https://github.com/grails/grails-core/issues/10451 due to which event handlers such as beforeInsert()/beforeUpdate() are
+                    /* Workaround to bug XXX due to which event handlers such as beforeInsert()/beforeUpdate() are
```
PiotrZakrzewski: Why is it replaced with XXX now? Is it a mistake during rebasing? I thought that it was first set to XXX and later on Jan put a link when he created an issue.

transmart-server/grails-app/conf/logback.groovy
```diff
@@ -0,0 +1,90 @@
...
+            fileNamePattern = "${logDirectory}/transmart.%i.log"
+            minIndex = 1
+            maxIndex = 9
+        }
+    }
+    root(WARN, ['transmart'])
+}
+
+/**
+ * Configuration for writing audit metrics.
```
PiotrZakrzewski: Is it still necessary to keep this uncommented code block?

transmart-oauth/grails-app/services/org/transmart/oauth/OAuth2SyncService.groovy
JanKanis: Yup, looks like a rebasing mistake

JanKanis:
modulo Piotr's comments i'm fine with merging this. However I'm out of office the next 3 weeks so someone else will have to hit 'merge'

transmart-server/grails-app/conf/logback.groovy
ewelinagr: This block is a part of audit metrics recording configuration, described on [transmartfoundation wiki](https://wiki.transmartfoundation.org/display/transmartwiki/Recording+audit+metrics). Originally it was a part of  
[Config.groovy](https://github.com/transmart/transmartApp/blob/master/grails-app/conf/Config.groovy#L206,) (also commented out by default)
but since Log4j has been replaced by grails-app/conf/logback.groovy for logging, it was moved here. I am not sure if it will be still used (outside of transmartApp), so I did not remove it, just in case.

transmart-oauth/src/main/groovy/org/transmart/oauth/TransmartOauthGrailsPlugin.groovy
ewelinagr: As I checked - bruteForceLoginLockService is properly injected in transmartApp, including settings for allowedNumberOfAttempts and lockTimeInMinutes (I tested it in LoginController)

transmartApp/grails-app/init/org/transmartproject/app/BootStrap.groovy
ewelinagr: According to [Grails documentation](http://docs.grails.org/latest/guide/plugins.html#providingBasicArtefacts) Bootstrap.groovy is excluded by default, when packaging a plugin. I moved the code from init() to the plugin descriptor closure: TransmartAppGrailsPlugin.doWithApplicationContext. Same for tm-oauth. There is still some config duplication in tmApp and tm-server, but I am not sure if I can safely deduplicate everything.

PR #7, opened 2017-04-11
a PR with the changes from #6 but without the first commit, to see the diff

JanKanis:
do not merge


transmartApp/src/main/groovy/org/transmart/FederatedUserDetailsService.groovy
```diff
@@ -1,6 +1,5 @@
 package org.transmart
```
JanKanis: This class (and SAML authentication in general) should probably also be moved to transmart-oauth. But it would be difficult to test if it still works...

transmartApp/grails-app/init/org/transmartproject/app/BootStrap.groovy
```diff
@@ -50,11 +50,6 @@ class BootStrap {
 
         // force marshaller registrar initialization
         grailsApplication.mainContext.getBean 'marshallerRegistrarService'
-
-        if ('clientCredentialsAuthenticationProvider' in
-                grailsApplication.config.grails.plugin.springsecurity.providerNames) {
-            OAuth2SyncService.syncOAuth2Clients()
-        }
     }
```
JanKanis: The blocks about SAML and configFine probably also should be in transmart-server. I don't know what the purpose of the forceEagerSessionCreation and concurrent sessions stuff is, so I'm not sure if it belongs here or in -server.

JanKanis: Wait, the configFine check is already there in tm-server, duplicated. Todo: deduplicate it.

JanKanis: And tm-server Bootstrap.groovy and tm-oauth Bootstrap.groovy are exact copies. That also needs deduplication. I think tm-oauth doesn't need most of what's in there.

gijskant:
#6 has been merged.

PR #9, opened 2017-04-14
Update database schema to load SNP data

ewelinagr:
- add missing oracle procedures
- add assay_id column to de_snp_subject_sorted_def table
- update delta scripts


rnugraha:
@ewelinagr I don't see update SQL for Postgres version. Is this out of scope now? Other than this I'm fine. @forus might want to have a look as well.

forus:
@rnugraha @ewelinagr looks fine for me. Porting to oracle is out of scope.

PR #10, opened 2017-04-19
Categorical values aggregate

JanKanis:
Add a 'values' aggregate that finds the distinct values in a categorical variable

bonus:
- Only do sanity checks if the main aggregate query fails, that saves us 2-3 (possibly expensive) database queries
- Allow multiple aggregates in a single request, that is much more efficient if the query takes a long time
- remove AggregateType.NONE which we don't need

Closes GDB-47

### DIFF
--- a/transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
+++ b/transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
@@ -10,6 +10,7 @@ import grails.util.Holders
 import groovy.transform.CompileStatic
 import groovy.transform.TupleConstructor
 import org.apache.commons.lang.NotImplementedException
+import org.hibernate.Criteria
 import org.hibernate.ScrollMode
 import org.hibernate.ScrollableResults
 import org.hibernate.SessionFactory
@@ -39,7 +40,7 @@ import org.transmartproject.core.multidimquery.Dimension
 import org.transmartproject.core.multidimquery.Hypercube
 import org.transmartproject.core.multidimquery.MultiDimConstraint
 import org.transmartproject.core.multidimquery.MultiDimensionalDataResource
-import org.transmartproject.core.multidimquery.MultiDimensionalDataResource.RequestConstraintsAndVersion
+import org.transmartproject.core.multidimquery.MultiDimensionalDataResource.RequestConstraintAndVersion
 import org.transmartproject.core.ontology.ConceptsResource
 import org.transmartproject.core.ontology.MDStudy
 import org.transmartproject.core.querytool.QueryResult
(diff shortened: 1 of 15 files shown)


transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -592,69 +647,110 @@ class MultidimensionalDataResourceService implements MultiDimensionalDataResourc
...
-                operator: Operator.AND,
-                args: [conceptConstraint, notNumericalCombination]
-        )
 
-        if (exists(builder, conceptNotNumericalCombination)) {
-            def message = 'One of the observationFacts had either an empty numerical value or a ' +
-                    'textValue with something else then \'E\''
-            throw new InvalidQueryException(message)
-        }
+        // Transmart currently does not allow null values in the ObservationFacts, but it could be extended to allow
```
gijskant: Why does it not allow that? Because of the exception that is thrown in `ObservationFact` or is there some deeper reason?

gijskant: Is the reason just so that no expensive checks are required for every aggregate call?

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/QueryController.groovy
```diff
@@ -153,19 +153,30 @@ class QueryController extends AbstractQueryController {
...
 
-        if (!args.type) {
+        if (!type) {
             throw new InvalidArgumentsException("Type parameter is missing.")
         }
+        if (!(type instanceof String || type instanceof List)) throw new InvalidArgumentsException(
+                "invalid type parameter (not a string or a list of strings)")
+
+        if (type instanceof String) {
+            type = [type]
```
gijskant: Dynamic types :)

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
JanKanis: The database and hypercube code could allow null values in ObservationFact by basically removing the checks like this one. However the tabular results will insert nulls for absent observations, so the old core db does not distinguish between nulls and absent values. Therefore the HddTabularResultHypercubeAdapter treats nulls in a tabular result as absent observations. 

I asked some data people, and as far as they were aware null values are not used and also not supported by kettle. So it seemed best to me to also disallow null values in the hypercube, hence the exception in ObservationFact. I just added a comment to document this.

I changed the aggregate implementation to first just try to retrieve results and only do the checks if no result is found. I don't like the doing 3 potentially expensive queries when 1 would suffice. This changes the semantics a bit, since if there are both valid values and invalid values it will now give an answer based on the valid values while previously it would return an error.

JanKanis:
AFAIK Transmart already doesn't really support null values. The tabular core-db treats null values and absent values the same, and apparently Kettle at least has no support for loading null values. Until now the clinical hypercube code did support nulls, but hdd never did as it is based on tabular results.

My guess is that the legacy transmart code will also break if it is given null values, based on the general quality of the code.

PR #16, opened 2017-05-11
Inline selections

JanKanis:
Inline patient set selections, so Peter and Bo can start working on it and (hopefully) have something working by tuesday.


JanKanis:
The transmart-batch tests are failing, but on master too. Probably due to an environment update done by travis. Gijs is looking into that, I think this PR can be merged if the other tests don't fail.

transmart-core-db/grails-app/domain/org/transmartproject/db/metadata/DimensionDescription.groovy
```diff
@@ -100,6 +100,6 @@ class DimensionDescription {
      * @return a list of all dimensions
      */
     static List<Dimension> getAllDimensions() {
-        findAll { name != LEGACY_MARKER }*.dimension
+        findAll { name != DimensionDescription.LEGACY_MARKER }*.dimension
```
ewelinagr: Is this change needed here? For other methods from this class that are using LEGACY_MARKER we do not use this kind of self-reference

transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/DimensionImpl.groovy
```diff
@@ -12,6 +12,7 @@ import groovy.transform.TupleConstructor
 import org.apache.commons.lang.NotImplementedException
 import org.grails.datastore.mapping.query.api.BuildableCriteria
 import org.hibernate.SessionFactory
+import org.hibernate.criterion.Projections
```
ewelinagr: Unused import

transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/query/Constraint.groovy
```diff
@@ -281,6 +283,12 @@ class FieldConstraint extends Constraint {
     Object value
 
     static constraints = {
+        field validator: { val, obj, Errors errors ->
```
ewelinagr: I think that [@BindUsing({ obj, source -> ConstraintFactory.bindField(obj, 'field', source['field']) })](https://github.com/thehyve/transmart-core/blob/8e6cba6583599dd474c09c39e77df594d9a3cf17/transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/query/Constraint.groovy#L279) is already checking if field is set, but I am not 100% sure.

transmart-core-db/grails-app/domain/org/transmartproject/db/metadata/DimensionDescription.groovy
JanKanis: I know, for some reason it is needed here, it fails without it. (though IntelliJ also thinks it is not needed)

transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/query/Constraint.groovy
JanKanis: I'm not sure what happens if the closure returns null, but at least if the relevant parameter ('field') is not present at all in the query, the @BindUsing won't be called and there won't be an error.

PR #23, opened 2017-05-18
Various fixes

JanKanis:
Several fixes. Each of the 3 commits is a separate issue, probably easier to review them one by one. See the commit messages.


transmart-rest-api/grails-app/controllers/org/transmartproject/rest/AbstractQueryController.groovy
```diff
@@ -75,25 +77,28 @@ abstract class AbstractQueryController implements Controller {
     }
 
     protected Constraint bindConstraint(constraintParam) {
-        Constraint constraint = getConstraintFromStringOrJson(constraintParam)
+        try {
+            return getConstraintFromStringOrJson(constraintParam)
+        } catch (ConstraintBindingException e) {
+            if(e.errors?.fieldErrors?.any()) {
```
forus: Isn't `.any()` redundant here?

forus: Shouldn't we process the rest of errors (non-field) as well here?

transmart-core-db-tests/build.gradle
```diff
@@ -27,4 +27,7 @@ assets {
     packagePlugin = true
 }
 
+// Force selected tests to always run, even if Gradle thinks they are up to date
```
forus: Is gradle is mistaken sometimes here or smth?

transmart-core-db-tests/src/main/groovy/org/transmartproject/db/StudyTestData.groovy
```diff
@@ -36,11 +36,15 @@ class StudyTestData {
                 if(!candidates) {
                     assert DimensionImpl.isBuiltinDimension(it), "Unknown dimension name '$it', " +
                             "modifier dimensions as string are not supported in createStudy()"
-                    study.addToDimensionDescriptions(new DimensionDescription(name: it))
+                    // saving is necessary here so that this dimension will be visible to other studies that are
+                    // created within this same transaction. Without it we will have duplicate dimensions.
+                    def dim = new DimensionDescription(name: it).save()
```
forus: Hm. I don't really understand this code.
Running below code for two different studies must try to produce 2 different dimension descriptions with the same name and then fail because of the constraint.
`new DimensionDescription(name: 'study').save()`

transmart-core-db-tests/build.gradle
JanKanis: Grails tries to only run tests whose code has changed, but with the TM packaging structure grails doesn't see correctly which tests should run again. This change makes grails always run all selected tests.

transmart-core-db-tests/src/main/groovy/org/transmartproject/db/StudyTestData.groovy
JanKanis: It does. The test code first creates a bunch of studies with their associated data, and then saves all of them at once. The `DimensionDescription.findAllByName(it)` does not find unsaved dimension descriptions, so each study gets its own set of dimension descriptions. When all the studies are saved this results in a uniqueness violation.

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/AbstractQueryController.groovy
JanKanis: Indeed, I updated the code

PR #33, opened 2017-05-29
Add API call that enables discovering elements of specified dimension

ewelinagr:
{{url}}/v2/dimension/{{dimensionName}}
Implement discovering of:
- trial visits
- visits
- non-builtin dimensions (modifiers)

GBD-17, GBD-49, GBD-69


open-api/swagger.json
```diff
@@ -1590,6 +1590,30 @@
           }
         }
       }
+    },
+    "/v2/dimension/{$dimensionName}": {
+      "get": {
+        "description": "Gets one patient object.\n",
```
JanKanis: Description doesn't look correct

open-api/swagger.yaml
```diff
@@ -1374,6 +1374,23 @@ paths:
       responses:
         204:
           description: 'returns null'
+  /v2/dimension/{$dimensionName}:
+    get:
+      description: |
+        Gets one patient object.
```
JanKanis: Description doesn't look correct

JanKanis:
I would suggest to name the endpoint `/dimensions/{dimensionName}/elements`. Then `/dimensions` can be used to retrieve which dimensions exist, `/dimensions/{dimensionName}` for information about the dimension without retrieving all elements. But those can be implemented later.

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -116,6 +116,13 @@ class MultidimensionalDataResourceService implements MultiDimensionalDataResourc
     Dimension getDimension(String name) {
         DimensionDescription.findByName(name).dimension
     }
+    
+    DimensionImpl getDimensionForStudies(String dimensionName, Collection<MDStudy> studies) {
```
JanKanis: Looking at this again, there is still a need for this call. But currently this call does two things: convert a string to a dimension object, and check that the the dimension is part of the accessible studies. It would be more clear to split those operations. Convert the string to dimension separately and then verify that the user can access them. So maybe you can change this to something like `Collection<Dimension> inaccessibleDimensions(Collection<Dimension> dimensions, User user)` that returns the dimensions to which the user does not have access and a wrapper `void verifyDimensionsAccessible(Collection<Dimension>, User user)` and move it them `AccessControlChecks`.

ewelinagr:
@JanKanis I added a few changes for this PR, can you have a look again?

JanKanis:
I made a bunch of changes, so I'm fine with the implementation now. Still missing: core-db-tests for each implementation to get elements/element counts. There are separate implementations for modifiers, for Visit, for compound and for simple i2b2 dimensions for both selectDimensionElements and elementCount

ewelinagr:
@JanKanis `DimensionElementSerializer` class is missing now, is it something that you created?

JanKanis:
All tests except TM-batch are working now. Someone should look into why that broke.

ewelinagr:
@JanKanis tests are added now, can we merge this pull request?

PR #37, opened 2017-06-07
Add rest call to retrieve all the saved patient sets accessible by the user

ewelinagr:
GET /v2/patient_sets
+ remove unused controller


transmart-rest-api-e2e/src/test/groovy/tests/rest/v2/PatientsSetSpec.groovy
```diff
@@ -169,4 +169,25 @@ class PatientsSetSpec extends RESTSpec {
...
+                acceptType: contentTypeForJSON,
+                statusCode: 400
+        ]
+
+        when: "I try to fetch all patientSets"
+        def responseData = get(request)
+
+        then: "the list of all patientSets is returned"
+        assert responseData.patientSets != null
+        responseData.patientSets.each {
```
JanKanis: This should also check that the list of patient sets is not empty. But to guarantee that the test also needs to create a patient set

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -584,6 +584,17 @@ class MultidimensionalDataResourceService implements MultiDimensionalDataResourc
         queryResult
     }
     
+    @Override List<QueryResult> findAllPatientSets(User user) {
+        List resultsWithUserAccess = []
+        List<QueryResult> queryResultList = QtQueryResultInstance.findAll()
```
JanKanis: It would be preferable if we can retrieve only the right query instances, instead of retrieving all of them and filtering. For non-admin users that would be retrieving the QtQueryResultInstance's by their queryInstance.userId. That does bypass the access control layer which is unfortunate, but I think it is preferrable to retrieving the potentially large set of query results.

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -582,6 +582,17 @@ class MultidimensionalDataResourceService implements MultiDimensionalDataResourc
         queryResult
     }
     
+    @Override List<QueryResult> findPatientSets(User user) {
+        List<QueryResult>  queryResultList = []
+        if(((DbUser) user).admin){
+            queryResultList = QtQueryResultInstance.findAll()
+        } else {
+            List<QtQueryInstance> queryInstanceList = QtQueryInstance.findAllByUserId(user.username)
+            queryResultList = QtQueryResultInstance.findAllByQueryInstanceInList(queryInstanceList)
```
JanKanis: If you do this with a subquery / criteria query it can be done in 1 query instead of two.

JanKanis:
I merged this PR so Bo can continue, but it still needs a core-db-tests test.

JanKanis:
Still missing: a core-db-tests test

PR #42, opened 2017-06-28
Add core-db-tests test for getPatientSets query

ewelinagr:
The test that was missing in [PR#37](https://github.com/thehyve/transmart-core/pull/37)


JanKanis:
Looks good, but if you are using constraint maps you migt as well make this a functional test instead of an integration test

ewelinagr:
Most of the tests from this class are using constraint maps, should they also be changed, or can it stay like this for now?

JanKanis:
Never mind, functional tests in core-db-tests don't work for v2 tests now because we can't load v2 data in that setup a.t.m.

PR #51, opened 2017-08-22
Fix tests around the dimensions test data.

forus:


transmart-core-db-tests/src/main/groovy/org/transmartproject/db/StudyTestData.groovy
```diff
@@ -21,37 +19,17 @@ class StudyTestData {
...
-                assert false, "Dimensions iterable passed to createStudy contains object that is not a String or a " +
-                        "DimensionDescription: $it"
-            }
-        }
-        study
+    static Study createStudy(String name, List<String> dimensions = [], boolean isPublic = false) {
+        new Study(
+            studyId: name,
+            secureObjectToken: isPublic ? Study.PUBLIC : "EXP:${name}",
+            dimensionDescriptions: DimensionDescription.findAllByNameInList(dimensions)
```
JanKanis: I think this needs a check that all requested dimensions are actually found

JanKanis:
Apart from that comment it looks fine

transmart-core-db-tests/src/main/groovy/org/transmartproject/db/StudyTestData.groovy
forus: Done

forus:
The code will try to fill the dictionaries on the production run as well.

PR #55, opened 2017-09-11
Fix generating tree node entries for non-categorical variables

ewelinagr:
Generating a missing parent node (a new row in the i2b2 table) was
possible only for categorical variables. Extended for other concept
types.

Fix for NOS-3: Numerical variables do not show up in the tree.

### DIFF
--- a/transmart-batch/src/main/groovy/org/transmartproject/batch/concept/ConceptTree.groovy
+++ b/transmart-batch/src/main/groovy/org/transmartproject/batch/concept/ConceptTree.groovy
@@ -115,6 +115,12 @@ class ConceptTree {
         if (node) {
             return node
         }
+        if(ClinicalVariable.conceptTypeFor(variable) != ConceptType.CATEGORICAL) {
+            /* to generate missing parent tree nodes for non-categorical variables */
+            for (def p = variable.path.parent; p != null; p = p.parent) {
+                getOrGenerate(p, null, ConceptType.UNKNOWN)
+            }
+        }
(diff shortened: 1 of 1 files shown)


transmart-batch/src/main/groovy/org/transmartproject/batch/concept/ConceptTree.groovy
```diff
@@ -115,6 +115,12 @@ class ConceptTree {
         if (node) {
             return node
         }
+        if(ClinicalVariable.conceptTypeFor(variable) != ConceptType.CATEGORICAL) {
```
gijskant: I think that generating missing parent nodes is also okay for categorical variables, even though they would have been added later by `getOrGenerateConceptNode` in `ClinicalFactsRowSetFactory`, especially because that only happens for 'non-strict' categorical nodes, meaning categorical nodes with value nodes below them.
So, the `if` is not needed here.

transmart-batch/src/main/groovy/org/transmartproject/batch/concept/ConceptTree.groovy
```diff
@@ -115,6 +115,12 @@ class ConceptTree {
...
+            /* to generate missing parent tree nodes for non-categorical variables */
+            for (def p = variable.path.parent; p != null; p = p.parent) {
```
gijskant: `getOrGenerate` already contains such a loop, so I think it is okay to just call the function on `variable.path.parent` (if not null).

PR #57, opened 2017-09-13
Add endpoint for fetching counts per concept.

gijskant:


transmart-rest-api/src/main/groovy/org/transmartproject/rest/marshallers/TreeNodeMarshaller.groovy
```diff
@@ -45,8 +50,14 @@ class TreeNodeMarshaller implements ObjectMarshaller<JSON> {
         if (constraint) {
             result.constraint = constraint
         }
+        if (obj.study?.studyId) {
+            result.studyId = obj.study?.studyId
```
ewelinagr: Guess the '?' mark is not needed inside if?

PR #62, opened 2017-10-04
Clear tree node cache

gijskant:
- Added endpoint for clearing the tree node cache
- Refactored tree node code
- Changed logic to fetch study id for a study node: dimcode if table is 'study', sourcesystemCd if table is 'concept_dimension'.

### DIFF
--- a/transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeNode.groovy
+++ b/transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeNode.groovy
@@ -1,0 +1,128 @@
+package org.transmartproject.core.tree
+
+import org.transmartproject.core.ontology.OntologyTerm
+import org.transmartproject.core.ontology.OntologyTermTag
+import org.transmartproject.core.ontology.OntologyTermType
+
+/**
+ * The ontology node class that represents nodes in the ontology forest.
+ * (There may be multiple root nodes, hence the name 'forest'.)
+ * Intermediate nodes contain a collection of child nodes.
+ * Leaf nodes represent a concept (either variable or value) or a study in TranSMART.
+ */
+interface TreeNode {
+
+    /**
+     * API version string used for marshalling the object.
+     * @return the API version string.
+     */
+    String getApiVersion()
+
+    void setApiVersion(String version)
+
+    /**
+     * The domain object on which this node is based.
+     */
+    OntologyTerm getDelegate()
+
+    /**
+     * Retrieves the parent node if it exists (i.e., when it is not a root node).
+     * @return the parent node.
+     */
+    TreeNode getParent()
+
+    void setParent(TreeNode parent)
+
+    /**
+     * Retrieves the child nodes.
+     * @return the child nodes.
+     */
+    List<TreeNode> getChildren()
+
+    /**
+     * Retrieves the tags associated with the node.
+     * @return the list of tags.
+     */
+    List<OntologyTermTag> getTags()
+
+    void setTags(List<OntologyTermTag> tags)
+
+    /**
+     * The number of observations that satisfy the constraint in {@link #getConstraint()}.
+     * The count is not always set, it can also be null.
+     * @return the number of observations for this node if set; null otherwise.
+     */
+    Long getObservationCount()
+
+    void setObservationCount(Long count)
+
+    /**
+     * The number of patients that have observations that satisfy the constraint in {@link #getConstraint()}.
+     * The count is not always set, it can also be null.
+     * @return the number of patients for this node if set; null otherwise.
+     */
+    Long getPatientCount()
+
+    void setPatientCount(Long count)
+
+    /**
+     * The name of the node.
+     * @return the name.
+     */
+    String getName()
+
+    /**
+     * The full tree path that contains all ancestors from the root up to the current node and the name of the node.
+     * @return the path.
+     */
+    String getFullName()
+
+    /**
+     * The visual attributes that characterise how the node should be represented.
+     * @return the set of visual attributes.
+     */
+    EnumSet<OntologyTerm.VisualAttributes> getVisualAttributes()
+
+    /**
+     * The dimension that the current node refers to. Could be study or concept.
+     */
+    String getDimension()
+
+    /**
+     * Name of the table the current node refers to. E.g., 'concept_dimension'.
+     */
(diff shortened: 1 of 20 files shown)


transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeNode.groovy
```diff
@@ -0,0 +1,128 @@
...
+     */
+    List<TreeNode> getChildren()
+
+    /**
+     * Retrieves the tags associated with the node.
+     * @return the list of tags.
+     */
+    List<OntologyTermTag> getTags()
+
+    void setTags(List<OntologyTermTag> tags)
```
forus: We don't have code outside core-db modifying tree nodes. Shouldn't we make this interface contains getters only?

transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeNode.groovy
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * The visual attributes that characterise how the node should be represented.
+     * @return the set of visual attributes.
+     */
+    EnumSet<OntologyTerm.VisualAttributes> getVisualAttributes()
+
+    /**
+     * The dimension that the current node refers to. Could be study or concept.
+     */
+    String getDimension()
```
forus: Why we repeat ontology term field in here? We could always access them via `delegate` field. BTW maybe we should come up with a better name in the core api for this field.

transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeNode.groovy
```diff
@@ -0,0 +1,128 @@
...
+    OntologyTermType getOntologyTermType()
+
+    /**
+     * Create a constraint object for the query that this node represents.
+     *
+     * Supports study, concept, modifier, patient and trial visit nodes.
+     *
+     * @return the constraint map for this node.
+     */
+    Map getConstraint()
```
forus: The implementation looks logic heavy. Wouldn't it be better to move it to a service? `Map getConstraint(OntologyTerm)`

transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeResource.groovy
```diff
@@ -0,0 +1,34 @@
+package org.transmartproject.core.tree
+
+import org.transmartproject.core.users.User
+
+interface TreeResource {
+
+    /**
+     * Finds tree nodes with prefix $rootKey up to depth $depth lower than the $rootKey level.
```
forus: It would be nice to mention that tree could vary from user to user and why.

transmart-core-db/grails-app/services/org/transmartproject/db/tree/TreeCacheService.groovy
```diff
@@ -95,7 +95,7 @@ class TreeCacheService {
                 node.children?.each { child ->
                     child.parent = node
                 }
-                node
+                node as TreeNode
```
forus: does it require explicit type casting?

transmart-core-db/src/main/groovy/org/transmartproject/db/tree/TreeNodeImpl.groovy
```diff
@@ -69,13 +70,6 @@ class TreeNode {
         delegate.columnName.toLowerCase().trim()
     }
```
forus: Why docs were removed?

transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeNode.groovy
gijskant: It is used outside the boundaries of the TreeResource.

transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeNode.groovy
gijskant: In this case, because we need to be able to retrieve the tags for a tree node and the tagsResource contract is based on OntologyTerm objects.
We could change that one of course, to create a `Map<String, List<OntologyTermTag>> getTags(Set<String> ontologyTermPaths, boolean includeDescendantsTags)`. I will look into it.

transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeNode.groovy
gijskant: Yes, it could be in a separate class. Although a service does not seem the right place, and I am tempted to make it return a proper constraint: `Constraint getConstraint(TreeNode)`.

transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeResource.groovy
gijskant: Agree. I will extend the documentation.

transmart-core-db/grails-app/services/org/transmartproject/db/tree/TreeCacheService.groovy
gijskant: Yes, it gave an error without it.

transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeNode.groovy
forus: @gijskant Where is it used then? I see usage in `TreeService` only which implements TreeResource. Anyway is there reason to keep the logic that sets the field outside the core?

transmart-core-db/grails-app/services/org/transmartproject/db/tree/TreeCacheService.groovy
forus: ah, it must be @CompileStatic

transmart-core-api/src/main/groovy/org/transmartproject/core/tree/TreeNode.groovy
gijskant: Fixed.

gijskant:
@forus Changes made according to feedback. Rebased on master.

PR #63, opened 2017-10-04
Fix aggregate combining

JanKanis:
This change should fix NTRDEV-133, though I haven't tested it yet. Still to be done: write a test to check the combinations of aggregates.

I changed the code to now accept any combination of aggregates, if the aggregates can not be retrieved in a single database query the code will split it into multiple queries.


JanKanis:
@forus I don't have time now to write a unit test, maybe you can do that. I will probably only have time to do it next week.

PR #66, opened 2017-10-12
Refactor aggregate call

forus:
Functional changes:
- Fix bug where `/aggregate?type=count&...` call fails.
- Count categorical values
- Standard deviation support

@beaugogh @gijskant These changes break the contract with the frontend. 
`/count_categorical_values` has to be used instead of `/aggregate?type=values&...` for the categorical values autocomplete.


transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -314,69 +314,28 @@ class MultidimensionalDataResourceService implements MultiDimensionalDataResourc
...
+            case AggregateFunction.COUNT:
                 return Projections.rowCount()
-            case AggregateType.PATIENT_COUNT:
+            case AggregateFunction.PATIENT_COUNT:
                 return Projections.countDistinct('patient')
-            case AggregateType.VALUES:
-                return Projections.distinct(Projections.property('textValue'))
+            case AggregateFunction.STD_DEV:
+                return Projections.sqlProjection(
+                        'STDDEV(nval_num) as SD',
```
gijskant: `STDDEV` or `STDDEV_SAMP`? In PostreSQL, stddev is an alias for stddev_samp (https://www.postgresql.org/docs/9.1/static/functions-aggregate.html), in Oracle, the differ (https://docs.oracle.com/cd/B19306_01/server.102/b14200/functions159.htm).

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -787,112 +746,44 @@ class MultidimensionalDataResourceService implements MultiDimensionalDataResourc
...
-            }
-        }
+        def builder = getCheckedQueryBuilder(user)
+        DetachedCriteria criteria = builder.buildCriteria(constraint)
+        def projections = Projections.projectionList()
+        projections.add(Projections.groupProperty('textValue'))
+        projections.add(Projections.rowCount())
+        criteria.setProjection(projections)
+                .setResultTransformer(Transformers.TO_LIST)
+                .add(Restrictions.eq('valueType', ObservationFact.TYPE_TEXT))
```
gijskant: Also filter out null values?

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -787,112 +746,44 @@ class MultidimensionalDataResourceService implements MultiDimensionalDataResourc
...
+        def projections = Projections.projectionList()
+        types.each { AggregateFunction aggregateType ->
+            projections.add(projectionForAggregate(aggregateType), aggregateType.toString())
         }
-
-        return result
+        criteria
+                .setProjection(projections)
+                .setResultTransformer(Transformers.ALIAS_TO_ENTITY_MAP)
+                .add(Restrictions.eq('valueType', ObservationFact.TYPE_NUMBER))
```
gijskant: Also filter out null values?

transmart-core-db/src/integration-test/groovy/org/transmartproject/db/multidimquery/QueryServicePgSpec.groovy
```diff
@@ -619,4 +607,77 @@ class QueryServicePgSpec extends Specification {
...
+                path: '\\Public Studies\\CATEGORICAL_VALUES\\Demography\\Race\\')
+        when:
+        def categoricalValueFreqMap = multiDimService.countCategoricalValues(race, user)
+        then:
+        categoricalValueFreqMap == [ Caucasian: 2L, Latino: 1L ]
+
+        when: 'only numerical observations selected'
+        def emptyMap = multiDimService.countCategoricalValues(new ConceptConstraint(
+                path: '\\Public Studies\\EHR\\Vital Signs\\Heart Rate\\'), user)
+        then: 'map is empty'
```
gijskant: I am not sure if I expect an error or an empty map here.

gijskant: What happens when doing a numerical aggregate on a categorical variable? Also empty map, right?

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
forus: Nice catch, thanks! I think we should use `STDDEV_SAMP` then instead.

transmart-core-db/src/integration-test/groovy/org/transmartproject/db/multidimquery/QueryServicePgSpec.groovy
forus: yes

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
forus: We are going to filter null values, but not null aggregate values. e.g. `{ stddev: null }` means there are less than two numerical values.

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
forus: It might be useful to know a number of missed values. We decided to keep them. `{null: 343}`. I will add test case for that.

PR #73, opened 2017-10-17
Add counts endpoint for observation and patient counts.

gijskant:


transmart-rest-api-e2e/src/test/groovy/tests/rest/v2/ObservationCountsSpec.groovy
```diff
@@ -33,6 +33,27 @@ class ObservationCountsSpec extends RESTSpec {
     }
 
     /**
+     *  given: "Study SHARED_CONCEPTS_RESTRICTED_LOADED is loaded, and I do not have access"
+     *  when: "I count observations for the concept Heart Rate"
+     *  then: "I get a count including observations from the restricted study"
+     */
+    def "restricted count2"() {
```
forus: `counts`? instead `count2`

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/QueryController.groovy
```diff
@@ -136,6 +139,27 @@ class QueryController extends AbstractQueryController {
...
+     *
+     * @return a the number of observations that satisfy the constraint and the number of associated patients.
+     */
+    def counts() {
+        def args = getGetOrPostParams()
+        checkForUnsupportedParams(args, ['constraint'])
+
+        Constraint constraint = bindConstraint(args.constraint)
+        if (constraint == null) {
+            return
```
forus: Why do we do this?

gijskant: On binding errors, `bindConstraint` will return null (and set response code 400).

PR #78, opened 2017-10-25
 Use concept code in tree node constraints. 

gijskant:


gijskant:
@forus I updated the PR to leave out the constraint labels.

PR #80, opened 2017-10-26
Json writer

gijskant:


transmart-rest-api/build.gradle
```diff
@@ -29,8 +29,8 @@ dependencies {
     testCompile "org.grails:grails-plugin-testing"
     testRuntime "net.sourceforge.htmlunit:htmlunit:2.18"
     testCompile 'org.grails:grails-datastore-rest-client'
-    compile 'org.quartz-scheduler:quartz:2.2.1'
     compile "com.opencsv:opencsv:${openCsvVersion}"
+    compile "org.grails.plugins:quartz:${quartzVersion}"
```
forus: Why you introduced this dependency for?

gijskant: Because it is required when running transmart-server without transmartApp. Not really related to the JSON writer, but I encountered this issue during testing (I was running without transmartApp for having a more light weight application).

transmart-rest-api/src/main/groovy/org/transmartproject/rest/serialization/HypercubeJsonSerializer.groovy
```diff
@@ -200,6 +159,52 @@ class HypercubeJsonSerializer extends HypercubeSerializer {
...
+            writer.value(value as String)
+        } else if (value instanceof Number) {
+            writer.value(value as Double)
+        } else if (value instanceof Date) {
+            def time = Instant.ofEpochMilli((value as Date).time).toString()
+            writer.value(time)
+        } else if (value instanceof Map) {
+            def obj = value as Map
+            writer.beginObject()
+            obj.each { k, v ->
```
JanKanis: @gijskant Closure calls can be quite slow, and in this form (probably) will be. In CompileStatic code you should use a for(Entry e : obj) { ... } loop.

transmart-rest-api/src/main/groovy/org/transmartproject/rest/serialization/HypercubeJsonSerializer.groovy
```diff
@@ -200,6 +159,52 @@ class HypercubeJsonSerializer extends HypercubeSerializer {
...
+        } else {
+            writer.value(value.toString())
+        }
+    }
+
+    protected void writeCell(Cell cell) {
+        writer.beginObject()
+        writer.name('inlineDimensions')
+        writer.beginArray()
+        cell.inlineDimensions.each {
```
JanKanis: idem

transmart-rest-api/src/main/groovy/org/transmartproject/rest/serialization/HypercubeJsonSerializer.groovy
```diff
@@ -200,6 +159,52 @@ class HypercubeJsonSerializer extends HypercubeSerializer {
...
+            writeValue(it)
+        }
+        writer.endArray()
+        writer.name('dimensionIndexes')
+        writer.beginArray()
+        cell.dimensionIndexes.each {
```
JanKanis: idem

transmart-rest-api/src/main/groovy/org/transmartproject/rest/serialization/HypercubeJsonSerializer.groovy
```diff
@@ -200,6 +159,52 @@ class HypercubeJsonSerializer extends HypercubeSerializer {
...
+            writer.value(it)
+        }
+        writer.endArray()
+        if (cell.numericValue != null) {
+            writer.name('numericValue').value(cell.numericValue)
+        } else if (cell.stringValue) {
+            writer.name('stringValue').value(cell.stringValue)
+        }
+        writer.endObject()
+        writer.flush()
```
JanKanis: Why do you need to flush here?

gijskant: Not sure if it's needed. I don't know how the memory usage is when not explicitly calling flush. I guess it can go.

PR #88, opened 2017-11-07
Tree nodes improvements

gijskant:


transmart-core-db/grails-app/services/org/transmartproject/db/tree/TreeService.groovy
```diff
@@ -181,7 +183,30 @@ class TreeService implements TreeResource {
...
-    final private Lock cacheRebuildLock = new ReentrantLock()
+    static class SimpleLock {
+        boolean locked
+    }
+
+    static final private SimpleLock sharedLock = new SimpleLock(locked: false)
+    static final private Lock lockLock = new ReentrantLock()
+
+    boolean tryLock() {
+        lockLock.lock()
```
forus: `lockLock`). Is it lock on lock?

transmart-core-db/grails-app/services/org/transmartproject/db/tree/TreeService.groovy
```diff
@@ -196,24 +221,42 @@ class TreeService implements TreeResource {
...
-        stopWatch.stop()
-        usersResource.getUsers().each { User user ->
-            log.info "Rebuilding the cache for user ${user.username} ..."
-            stopWatch.start("Rebuild the cache for ${user.username}")
-            treeCacheService.fetchCachedSubtree(user.admin, getStudyTokens(user), I2b2Secure.ROOT, 0)
-            stopWatch.stop()
-        }
-        log.info "Done rebuilding the cache.\n${stopWatch.prettyPrint()}"
-        cacheRebuildLock.unlock()
+    void rebuildCache(User currentUser) throws ServiceNotAvailableException {
```
forus: What was wrong with the old way? Didn't `Async` work?

open-api/swagger.yaml
```diff
@@ -1239,6 +1239,10 @@ paths:
           in: query
           type: integer
           description: 'The max node depth returned'
+        - name: constraints
+          in: query
+          type: boolean
+          description: 'Flag if the constraints should be included in the result (always true for hal, defaults to false for json)'
```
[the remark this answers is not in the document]
gijskant: The difference has to do with the different nature of HAL and JSON: JSON has all children embedded (of course, this can be limited by the depth parameter). If constraints are always included, that blows up the JSON object.
For HAL it is different, it does not include children, but links to children. The constraint objects are only provided for nodes at the top level.

transmart-core-db/grails-app/services/org/transmartproject/db/tree/TreeService.groovy
gijskant: No, it didn't :(
In normal Spring Boot it works, but Grails apparently has its own methods.
Probably we could make the Spring Boot way to work, but I don't know how.

transmart-core-db/grails-app/services/org/transmartproject/db/tree/TreeService.groovy
gijskant: Indeed. The `lockLock` can not be locked by one thread and unlocked by another. The sharedLock is used for that. The lockLock is to make sure that one thread can check or change the shared lock at the same time.
The scenario is here that the controller thread checks if a rebuild process is currently active. If not, it starts one and activates the lock at the same time.
This spawns a new task with its own thread. That thread has to unlock the lock at the end of the task.

gijskant: I see that I should add documentation in the code.

transmart-rest-api/src/main/groovy/org/transmartproject/rest/serialization/TreeJsonSerializer.groovy
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Writes the tree node to JSON.
+     *
+     * @param out the stream to write to.
+     */
+    void write(Map args, final TreeNode node, OutputStream out) {
+        this.writer = new JsonWriter(new PrintWriter(new BufferedOutputStream(out)))
+        this.writeConstaints = args?.writeConstraints ?: false
+        writeNode(node)
```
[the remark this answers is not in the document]
gijskant: Not `endObject`, because there is also no `beginObject` (that happend in `writeNode`), but `flush` may be a good idea.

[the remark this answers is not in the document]
gijskant: Not `endObject`, because there is also no `beginObject` (that happend in `writeNode`), but `flush` may be a good idea.

PR #89, opened 2017-11-10
Pedigree

gijskant:


transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/query/Constraint.groovy
```diff
@@ -635,6 +635,28 @@ class SubSelectionConstraint extends Constraint {
     }
 }
 
+@Canonical
+class RelationConstraint extends Constraint {
```
gijskant: @forus The rest of the API uses underscores, here it is camel case, we probably want to change this class to use underscores.

PR #97, opened 2017-11-23
Add export to SPSS .sav files.

gijskant:


transmart-rest-api/src/main/groovy/org/transmartproject/rest/dataExport/WorkingDirectory.groovy
```diff
@@ -0,0 +1,32 @@
...
+            jobTmpDirectory = tempFolderDirectory
+        }
+        String userDirectory = jobTmpDirectory + File.separator + user.username
+        File fileDirectory = new File(userDirectory)
+
+        if (!fileDirectory.isDirectory()) fileDirectory.mkdirs()
+        fileDirectory
+    }
+
+    static File createDirectoryUser(User user, String prefix, String suffix = null) {
```
forus: @gijskant Does it mean that we store only the last exported files per user?

PR #107, opened 2017-11-28
Serialise tree node tags when tags are requested.

gijskant:


transmart-rest-api/src/main/groovy/org/transmartproject/rest/serialization/TreeJsonSerializer.groovy
```diff
@@ -43,10 +57,14 @@ class TreeJsonSerializer {
         if (node.patientCount) {
             writer.name('patientCount').value(node.patientCount)
         }
-        if (this.writeConstaints && node.constraint) {
+        if (this.writeConstraints && node.constraint) {
             writer.name('constraint')
             writeConstraint(node.constraint)
         }
+        if (this.writeTags && node.tags) {
```
forus: @gijskant imho it's useful to have `metadata: {}` when the tags collection is empty.

PR #108, opened 2017-11-28
Remove duplicate FmFolderJob file

petcupaula:
The FmFolderJob.groovy appears twice (in transmartApp and then in folder-management-plugin). This results in an error when deploying transmart.war to Tomcat.


gijskant:
Thanks for the fix!

PR #110, opened 2017-11-28
Update dependency path to dataExportRScripts

petcupaula:
The current path `src/main/resources/dataExportRScripts` doesn't exist. The folder is however available in transmartApp. Should we update the path to transmartApp or move the scripts at the specified location?


gijskant:
In this case, the R scripts are a dependency for transmartApp specifically, so this seems to be the correct fix. Probably, we should do a similar change for `src/main/resources/public`.

PR #111, opened 2017-11-28
Cache counts per concept for better performance.

forus:


gijskant:
@forus Maybe also add a line in the `rebuildCache` call?

forus:
@gijskant It's a good idea. Done.

PR #120, opened 2017-12-02
Make copy loader options more independent from each other.

forus:


transmart-copy/src/main/groovy/org/transmartproject/copy/Database.groovy
```diff
@@ -68,25 +65,19 @@ class Database {
...
             host = 'localhost'
         }
-        String port = getenv('PGPORT') ?: '5432'
-        String database = getenv('PGDATABASE') ?: 'transmart'
+        String port = params['PGPORT'] ?: '5432'
+        String database = params['PGDATABASE'] ?: 'transmart'
         String url = "jdbc:postgresql://${host}:${port}/${database}"
         String username
         String password
-        if (connectAsAdmin) {
```
gijskant: For use with the `vars` file from `transmart-data` it is important to distiguish the pairs `$PGUSER` with accompanying `$PGPASSWORD` and the user `tm_cz` with accompanying password `$TM_CZ_PWD` (defaults to `tm_cz`).

forus: I see. It looks like program required one execution with `--admin` flag and other without anyway. Now the same could be achieved by running the command with changed env. variables. `PGPASSWORD=tm_cz PGPASSWORD =TM_CZ_PWD <comand>`

PR #122, opened 2017-12-07
Adjust transmart-batch to load high-dim data in the new way

ewelinagr:
Fix NOS-4 - loading of gene expression data

Make tm-batch load high-dim data with the new observations/modifiers
that this data needs in 17.1:
- High dim observations having textual value with sample name (subject
sample mapping).
- Modifier on it specifying high dim data type and as value assay ID.


transmart-batch/src/main/groovy/org/transmartproject/batch/facts/DummyFactGenerator.groovy
```diff
@@ -50,15 +64,39 @@ class DummyFactGenerator extends AbstractItemCountingItemStreamItemReader<Clinic
...
+
+            List<Map> result = getAssayIds(sample_id)
+
+            if (result.empty) {
+                return
+            }
+            // add modifier for each assay_id
+            String modifier = 'TRANSMART:HIGHDIM:' + platform.markerType.toUpperCase()
+            for (int i=0; i<result.size(); i++){
+                rowSet.instanceNum = i+1
```
forus: it does not look right. `instance_id` should be unique for each row. Pass it maybe to `addValue` method?

transmart-batch/src/main/groovy/org/transmartproject/batch/facts/DummyFactGenerator.groovy
```diff
@@ -72,4 +110,30 @@ class DummyFactGenerator extends AbstractItemCountingItemStreamItemReader<Clinic
...
+
+    TrialVisit getTrialVisit() {
+        def trialVisit = jdbcTemplate.queryForList """
+                SELECT *
+                FROM $Tables.TRIAL_VISIT_DIMENSION
+                WHERE study_num = :study_num
+            """, [study_num: study.studyNum]
+
+        if (trialVisit) {
+            def t = trialVisit?.first()
```
forus: Would it work for the data people to bind row with a first trial visit from the study? If so you should do it once per job.

transmart-batch/src/main/groovy/org/transmartproject/batch/facts/DummyFactGenerator.groovy
```diff
@@ -72,4 +110,30 @@ class DummyFactGenerator extends AbstractItemCountingItemStreamItemReader<Clinic
 
     @Override
     protected void doClose() throws Exception {}
+
+    private List<Map<String, Object>> getAssayIds(String sample_id) {
+        List result = jdbcTemplate.queryForList """
+                SELECT DISTINCT assay_id
```
forus: why `DISTINCT`?

transmart-batch/src/main/groovy/org/transmartproject/batch/facts/DummyFactGenerator.groovy
```diff
@@ -72,4 +110,30 @@ class DummyFactGenerator extends AbstractItemCountingItemStreamItemReader<Clinic
 
     @Override
     protected void doClose() throws Exception {}
+
+    private List<Map<String, Object>> getAssayIds(String sample_id) {
+        List result = jdbcTemplate.queryForList """
```
forus: What about making the result of the query the main data source, instead of reading from `subjectConceptMappingFileRows`?

transmart-batch/src/main/groovy/org/transmartproject/batch/facts/DummyFactGenerator.groovy
ewelinagr: I wasn't sure how to get a trial visit there and what to do, if there is more than one visit or none (for now there is a function to add a default visit then)

PR #123, opened 2017-12-07
Fix issues related to running on Windows

dennyverbeeck:
- Add a `pathingJar` for transmart-server to deal with long classpaths
- Replace backslashes in `user.home` system property with forward slashes so it can be used as part of a URI

### DIFF
--- a/transmart-server/build.gradle
+++ b/transmart-server/build.gradle
@@ -40,14 +40,45 @@ assets {
     packagePlugin = true
 }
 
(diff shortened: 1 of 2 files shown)


gijskant:
Thanks for the fix. Hopefully, there will be support for Windows on Travis at some point, so that we could run the tests easily on that platform as well.
I noticed that I had to increase the java heap space for gradle to compile. Did you have similar issues?

PR #128, opened 2017-12-14
Reuse patient sets

forus:
Based on optimizations to [the patient set creation functionality](https://github.com/thehyve/transmart-core/pull/127).


gijskant:
This change is included in #155.

PR #129, opened 2017-12-14
Don't limit on studies if user is admin.

forus:


forus:
@gijskant Thanks for reviewing. I've made one more change could you review [the last commit](https://github.com/thehyve/transmart-core/pull/129/commits/e4b57af3c9ff723df1e24b5609097ed014222952). Thanks!

PR #138, opened 2017-12-29
Query subscription

ewelinagr:


ewelinagr:
TODO: 
- extend tests
- change an email message (QueryDiffSubscriptionService)

transmart-server/src/main/groovy/org/transmart/server/subsctiption/SubscriptionFrequency.groovy
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * The frequency of emails for user query subscription
+ */
+@CompileStatic
+enum SubscriptionFrequency {
+
+    DAILY ('DAILY'),
+    WEEKLY ('WEEKLY')
+
+    private String subscriptionFrequency
```
forus: Enum names match exactly with their `subscriptionFrequency` values. I don't see a point in having this field so far.

transmart-server/src/main/groovy/org/transmart/server/subsctiption/ChangeFlag.groovy
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * The update data change type
+ */
+@CompileStatic
+enum ChangeFlag {
+
+    ADDED ('ADDED'),
+    REMOVED ('REMOVED')
+
+    private String changeFlag
```
forus: see comment for `SubscriptionFrequency`

transmart-server/grails-app/services/org/transmart/server/subscription/QueryDiffSubscriptionService.groovy
```diff
@@ -0,0 +1,127 @@
...
+ * Generates and sends a daily or weekly subscription email for each user
+ * that subscribed for a query data result updates.
+ *
+ * The DSL for sending emails is provided by the Grails mail plugin.
+ * See Grails Mail Plugin documentation for more information:
+ * http://gpc.github.io/grails-mail/
+ *
+ */
+@Slf4j
+class QueryDiffSubscriptionService {
```
forus: `QueryDiffSubscriptionMailService` sounds more precise

open-api/swagger.yaml
```diff
@@ -1929,6 +1937,40 @@ paths:
       responses:
         204:
           description: 'returns null'
+  /v2/query_diffs/{queryId}:
```
gijskant: Wouldn't `/v2/queries/{queryId}/updates` make more sense here? They refer to the same resource `/v2/queries/{queryId}`.

transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/MultiDimensionalDataResource.groovy
```diff
@@ -102,7 +102,20 @@ interface MultiDimensionalDataResource {
...
+    /**
+     * Update existing patient set that satisfies provided constraints
+     * @param name
+     * @param constraint
+     * @param user
+     * @param constraintText
+     * @param apiVersion
+     * @return A patient set with updated results (patient ids)
+     */
+    QueryResult updatePatientSetQueryResult(String name, MultiDimConstraint constraint, User user, String constraintText)
```
gijskant: Seems that `constraint` and `constraintText` are duplicates. If it is an existing patient set (or query), also the id of the set/query can be used.

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -708,6 +767,25 @@ class MultidimensionalDataResourceService implements MultiDimensionalDataResourc
...
     }
 
+    @Override QueryResult findQueryResultByConstraint(String constraintText, User user) {
+        QtQueryMaster queryMaster = getQueryMasterByConstraint(user, constraintText)
+        if (queryMaster == null) {
+            log.info "Patient set not found with specified constraints."
+            return null
+        }
+        QtQueryInstance queryInstance = queryMaster.queryInstances.find {
+            it.startDate == queryMaster.queryInstances*.startDate.max() && it.deleteFlag == 'N'
```
gijskant: This looks like sorting.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryDiffService.groovy
```diff
@@ -0,0 +1,242 @@
...
+            log.info "No queries were found."
+            return
+        }
+
+        for (query in userQueries) {
+            DbUser queryUser = (DbUser) usersResource.getUserFromUsername(query.username)
+            def oldSet = getOldSet(query, queryUser)
+            if (!oldSet) {
+                log.info "Set for query: '$query.id' was not found."
+                return
```
gijskant: Why return? I would say this initial diff contains all entries.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryDiffService.groovy
```diff
@@ -0,0 +1,242 @@
...
+                Constraint patientConstraint = createConstraints(query.patientsQuery)
+                QueryResult newSet = multiDimService.updatePatientSetQueryResult(
+                        query.name, patientConstraint, user, JSON.parse(query.patientsQuery)?.constraint.toString())
+
+                createQueryDiffWithEntries(oldSet, newSet, query)
+            }
+        }
+    }
+
+    private Object getOldSet(UserQuery query, DbUser user) {
```
gijskant: Why Object?

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryDiffService.groovy
```diff
@@ -0,0 +1,242 @@
...
+                        changeFlag: 'REMOVED'
+                )
+            })
+            queryDiff.save(flush: true)
+            queryDiffEntries*.save(flush: true)
+        }
+    }
+
+    private static List<Long> getAddedIds(List<Long> oldPatientIds, List<Long> newPatientIds) {
+        return newPatientIds - oldPatientIds
```
gijskant: This function is longer than its body...

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryDiffService.groovy
```diff
@@ -0,0 +1,242 @@
...
+
+    private Object getOldSet(UserQuery query, DbUser user) {
+        //get the latest queryDiff entry for the current query 
+        QueryDiff latestQueryDiff = getLatestQueryDiffByQueryId(query.id)
+        if (latestQueryDiff) {
+            return findSetByIdAndType(latestQueryDiff.setId, latestQueryDiff.setType, user)
+        } else {
+            //todo different methods depending on query type (for now patient only)
+            try {
+                def constraint = JSON.parse(query.patientsQuery)?.constraint
```
gijskant: Can't this JSON parsing be moved to the UserQuery class instead?

transmart-core-db/src/main/groovy/org/transmartproject/db/userqueries/SetTypes.groovy
```diff
@@ -0,0 +1,29 @@
...
+
+    private String type
+
+    SetTypes(String type) {
+        this.type = type
+    }
+
+    static SetTypes from(String type) {
+        SetTypes t = values().find { it.type == type }
+        if (t == null) log.error "Unknown type of set: ${type}"
```
gijskant: Please use `{}`. Error or warn? It does return null.

transmart-data/ddl/oracle/biomart_user/query.sql
```diff
@@ -55,3 +62,5 @@ COMMENT ON COLUMN biomart_user.query.observations_query IS 'The observation sele
 COMMENT ON COLUMN biomart_user.query.api_version IS 'The version of the API the query was intended for.';
 COMMENT ON COLUMN biomart_user.query.bookmarked IS 'Flag to indicate if the user has bookmarked the query.';
 COMMENT ON COLUMN biomart_user.query.deleted IS 'Flag to indicate if the query has been deleted.';
+COMMENT ON COLUMN biomart_user.query.subscribed IS 'Flag to indicate if the user has subscribed to the query.';
+COMMENT ON COLUMN biomart_user.query.subscription_freq IS 'Frequency of query notifications: daily or weekly digest.';
```
gijskant: Please add exact names: 'DAILY' or 'WEEKLY'.

transmart-server/src/main/groovy/org/transmart/server/subsctiption/SubscriptionFrequency.groovy
```diff
@@ -0,0 +1,29 @@
...
+
+    SubscriptionFrequency(String subsciptionFrequency) {
```
gijskant: typo

transmart-server/src/main/groovy/org/transmart/server/subsctiption/ChangeFlag.groovy
```diff
@@ -0,0 +1,29 @@
...
+
+    ChangeFlag(String changeFlag) {
+        this.changeFlag = changeFlag
+    }
+
+    static ChangeFlag from(String changeFlag) {
+        ChangeFlag f = values().find { it.changeFlag == changeFlag }
+        if (f == null) throw new Exception("Unknown change flag: ${changeFlag}")
```
gijskant: `{}`

transmart-server/grails-app/services/org/transmart/server/subscription/QueryDiffSubscriptionService.groovy
```diff
@@ -0,0 +1,127 @@
...
+        def currentDate = new Date()
+
+
+        List<UserQueryDiffEntry> queryDiffEntries = userQueryDiffResource
+                .getAllEntriesByUsernameAndFrequency(frequency.value(), username, firstResult, numResults)
+        def queryDiffsMap = queryDiffEntries?.groupBy { it.queryDiff }
+
+        StringBuilder textStringBuilder = new StringBuilder()
+
+        textStringBuilder.append("Change log of ${currentDate.format("d.' of 'M Y h:mm aa z")}")
```
gijskant: Or use some template engine for this (just a suggestion).

transmart-server/grails-app/conf/application.groovy
```diff
@@ -323,6 +323,36 @@ grails { plugin { springsecurity {
 } } }
 /* }}} */
 
+/* {{{ GlowingBear query subscription configuration */
+grails {
+    mail {
+        host = "smtp.gmail.com"
```
gijskant: A more sensible default is `localhost:25`.

transmart-server/grails-app/conf/application.groovy
```diff
@@ -323,6 +323,36 @@ grails { plugin { springsecurity {
...
+        'default' {
+            from = "subscription.gb@gmail.com"
+        }
+        port = 465
+        username = "subscription.gb@gmail.com"
+        password = "gb@TheHyve"
```
gijskant: Please change credentials and remove this line here. You don't want people spamming from this account.

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
forus: what if a row with `deleteFlag == 'Y'` would have a maximum start date?

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryDiffService.groovy
forus: I find `set` name very confusing in java word. Wouldn't it be better to name it `getPreviousQueryResult` ?

forus:
@ewelinagr @gijskant 
General thoughts:
- If we don't have sample type yet just clean types fields from the code. It will make the code cleaner.
- I've mentioned this idea before. Let's revisit it again. What if we would get rid of the dependency on the i2b2 qt_* tables completely by having an initial version of the result in the diff entries tables.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryDiffService.groovy
ewelinagr: I wanted to extend this method later to return other set types, that is why I didn't make it QueryResult specific, but I can change it for now, since it is not determined yet how the subject set (and other types if any) will look like.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryDiffService.groovy
ewelinagr: I'm not sure, all new entries would be send in an email with daily/weekly updates and what I understood is that this should contain information only about sets that where changed because of the data change (after data loading). I'll ask for that to make sure.

open-api/swagger.yaml
ewelinagr: @gijskant True, so as v2/query_diffs/scan. Do you think it should be changed to /v2/queries/scan? Would it be better to remove UserQueryDiffController and move all its methods to UserQueryController then?

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
```diff
@@ -0,0 +1,267 @@
...
+
+            List<QuerySetInstance> previousQuerySetInstances = getInstancesForLatestQuerySet(query.id)
+            ArrayList<Long> newPatientIds = getPatientsForQuery(query, user)
+
+            if (createSetWithDiffEntries(previousQuerySetInstances*.objectId, newPatientIds, (Query) query)) {
+                numberOfResults++
+            }
+
+        }
+        return numberOfResults
```
gijskant: You are returning userQueries.size()

transmart-data/updatedb/dev/oracle/Makefile
```diff
@@ -2,6 +2,9 @@ include ../../../lib/makefile.inc
 
 FILES = \
 	../../../ddl/oracle/biomart_user/query
+	../../../ddl/oracle/biomart_user/query_set
+    ../../../ddl/oracle/biomart_user/query_set_instance
```
gijskant: `make` requires tabs.

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQueryController.groovy
```diff
@@ -38,41 +46,53 @@ class UserQueryController {
 
     def save(@RequestParam('api_version') String apiVersion) {
         def requestJson = request.JSON as Map
-        checkForUnsupportedParams(requestJson, ['name', 'patientsQuery', 'observationsQuery', 'bookmarked'])
+        checkForUnsupportedParams(requestJson, ['name', 'patientsQuery', 'observationsQuery', 'bookmarked',
```
gijskant: Why not just specify an object with these fields as parameter of the function?

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
ewelinagr: not exactly, I am returning a number of userQueries that were updated (so a number of newly created querySets)

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQuerySetController.groovy
```diff
@@ -0,0 +1,107 @@
...
+        int firstResult = params.firstResult ?: 0
+        def numResults = params.numResults as Integer
+
+        List<UserQuerySetDiff> setDiffs = userQuerySetResource.getDiffEntriesByQueryId(queryId, currentUser,
+                firstResult, numResults)
+        def queryDiffsMap = setDiffs?.groupBy { it.querySet }
+        respond([querySetDiffs: queryDiffsMap ? queryDiffsMap.collect { toResponseMap(it, UserQuerySetDiff.class) } : []])
+    }
+
+    private static Map<String, Object> toResponseMap(response, Class<?> classType) {
```
gijskant: Does does not seem right. This combines three different serialisers in one. And it uses custom code for a very generic problem, for which solutions exist.
The type of `response` is apparently `Map<UserQuery, Object>`, so let's add that explicitly. You could either add marshallers for `UserQuery`, `UserQuerySetDiff` and `UserQuerySetInstance` (see, e.g., `MDStudySerializationHelper`). If high performance is required, you could write a custom serialise, as in `TreeController`.

transmart-core-api/src/main/groovy/org/transmartproject/core/userquery/UserQuerySetDiff.groovy
```diff
@@ -0,0 +1,22 @@
...
+
+    /**
+     * The id of the object that was updated in the query set e.g. patient.id
+     */
+    Long getObjectId()
+
+    /**
+    *  Flag determining whether the object was added or removed
+    */
+    String getChangeFlag()
```
gijskant: The enum can be used here, right?

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQueryController.groovy
ewelinagr: I am not sure, this is the way a request body is checked and parsed in all the other controllers, I didn't want to refactor this now, just extended the existing way. But if you think it would be beneficial to change it, I can do it.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
```diff
@@ -0,0 +1,267 @@
...
+        if(numResults) {
+            criteria.setMaxResults(numResults)
+        }
+        def result = criteria.list()
+        if (!result) {
+            return []
+        }
+
+        DbUser user = (DbUser) usersResource.getUserFromUsername(currentUser.username)
+        if (!user.admin && result.query.first().username != currentUser.username) {
```
gijskant: This was already checked on line 1, right?

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQuerySetController.groovy
```diff
@@ -0,0 +1,107 @@
...
+     */
+    def scan() {
+        Integer result = userQuerySetResource.scan(currentUser)
+        response.status = 201
+        respond([numberOfUpdatedSets: result])
+    }
+
+    /**
+     * Gets a list of query sets related to a specific query:
+     * <code>/${apiVersion}/query_sets/${queryId}</code>
```
gijskant: Could be `/${apiVersion}/queries/${queryId}/query_sets`, as it refers to the same query as in the userquerycontroller.

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQuerySetController.groovy
```diff
@@ -0,0 +1,107 @@
...
+     * @param numResults - number of results - parameter required for pagination
+     * @param queryId - id of the query
+     * @return list of querySets
+     */
+    def getSetsByQueryId(@PathVariable('queryId') Long queryId) {
+        checkForUnsupportedParams(params, ['queryId', 'firstResult', 'numResults'])
+        int firstResult = params.firstResult ?: 0
+        def numResults = params.numResults as Integer
+
+        List<UserQuerySetInstance> querySetInstances = userQuerySetResource.getSetInstancesByQueryId(queryId, currentUser,
```
gijskant: Why not just fetch a user query set, which has a set of patients/patient ids or object ids? Something similar happens but then with lists of set entries that are grouped, but I think the contract of the service call should be different.

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQuerySetController.groovy
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Gets a list of query result change entries by query id - history of data sets changes for specific query:
+     * <code>/${apiVersion}/query_sets/diffs/${queryId}</code>
+     *
+     * @param firstResult - first result - parameter required for pagination, by default equals 0
+     * @param numResults - number of results - parameter required for pagination
+     * @param queryId - id of the query
+     * @return list of queryDiffs
+     */
```
gijskant: What is paginated here, the object ids, or the snapshots/diffs?

gijskant: These object ids are quite meaningless numbers (patient nums are generated by the database, automatically incremented).
- Two questions: why would you want to expose those in the first place?
- Do you want to send them every time you send the list of diffs, or would it be better to have a separate call to retrieve the patients that are added/removed in a specific diff, e.g, `/v2/queries/${queryId}/diffs/${diffId}/changes`.

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQueryController.groovy
gijskant: There is not that many places where we actually create/edit data in transmart. The [StorageSystemController](https://github.com/thehyve/transmart-core/blob/master/transmart-rest-api/grails-app/controllers/org/transmartproject/rest/StorageSystemController.groovy) is an example, but that uses the generic `RestfulController` class, that may not be the best choice here.
With so many parameters (that are actually properties of the user query object), it makes sense to wrap them either in a command object, or pass them directly as user query objects.

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQuerySetController.groovy
ewelinagr: When I tried to fetch the user query set directly, querySetInstances were empty (same for querySetDiffs) and I couldn't make it working, that is why I added a different solution with grouping.

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQuerySetController.groovy
ewelinagr: This one refers to a userQuerySetController, 
also the call for diffs is:
` /${apiVersion}/queries/${queryId}/query_sets/diffs`
That is why I decided to use `/query_sets/${queryId}`.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
ewelinagr: I removed 1st line, because it was blocking an access to all querySets for admin users .

transmart-core-api/src/main/groovy/org/transmartproject/core/userquery/UserQuerySetDiff.groovy
ewelinagr: ChangeFlag enum is a part of core-db, so there is no access from core-api, but I can move the enum class to a different module, if it is needed.

gijskant: Yes, please move it to core-api.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
gijskant: I don't understand. Line 84 has `userQueryService.get(queryId, currentUser)`, but nothing happens with the result.
And why would an admin want to fetch another user's query set?

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQuerySetController.groovy
gijskant: The controller does not determine the API call. You can distribute arbitrarily over controllers as you like. In this case fetching query objects in one controller and query sets in another one. But these query sets belong to the queries, so it seems natural to use `/queries/${queryId}` in the url. `/query_sets/${queryId}` suggests that the `queryId` identifies one query set.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
gijskant: Sorry, forgot to pull. I will look again.

gijskant: Question remains: why not just include a restriction on the username in the query? I don't see a use for admins to retrieve arbitrary sets.

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQuerySetController.groovy
gijskant: Well, but this is really hard to read. And what is the relation between `getSetsByQueryId` and  `getDiffsByQueryId`. What are their types? Etc.

gijskant: These wrappers are really not needed. The data showcase is a better example of how to write a controller than most of Transmart, so please don't be inspired by the other controllers here.

transmart-server/grails-app/services/org/transmart/server/subscription/QuerySetSubscriptionMailService.groovy
```diff
@@ -3,6 +3,7 @@ package org.transmart.server.subscription
 import grails.plugins.mail.MailService
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
+import org.transmartproject.core.userquery.UserQuerySetChangesRepresentation
 import org.transmartproject.core.userquery.UserQuerySetDiff
 import org.transmartproject.core.userquery.ChangeFlag
```
forus: Unused imports. You could use an IDE to find them in the rest of places.

transmart-server/grails-app/services/org/transmart/server/subscription/QuerySetSubscriptionMailService.groovy
```diff
@@ -3,6 +3,7 @@ package org.transmart.server.subscription
 import grails.plugins.mail.MailService
 import groovy.util.logging.Slf4j
```
forus: Sorry for noticing this only now. Why do we have this logic in `transmart-server`? Saying that I understand that neither `transmart-core-db`, nor `transmart-rest-api` are proper places. What about having a new module e.g. `transmart-mail`?

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQueryController.groovy
forus: @ewelinagr  It looks applicable to the new code. You might want to consider this.

transmart-server/grails-app/services/org/transmart/server/subscription/QuerySetSubscriptionMailService.groovy
```diff
@@ -0,0 +1,139 @@
...
+     * - over what period the change was
+     *
+     * @param username
+     * @param freq
+     * @return The body of the email
+     *
+     */
+    private String generateEmail(String username, SubscriptionFrequency frequency) {
+
+        Integer maxNumberOfSets = grailsApplication.config.org.transmart.server.subscription.maxNumberOfSets
```
forus: it could be a field of this service instead.

transmart-server/grails-app/services/org/transmart/server/subscription/QuerySetSubscriptionMailService.groovy
```diff
@@ -0,0 +1,139 @@
...
+        def currentDate = new Date()
+
+        List<UserQuerySetChangesRepresentation> querySetChanges = userQueryDiffResource
+                .getQueryChangeHistoryByUsernameAndFrequency(frequency, username, maxNumberOfSets)
+
+        if (querySetChanges.size() > 0) {
+
+            StringBuilder textStringBuilder = new StringBuilder()
```
forus: You might want to use groovy multiline Gstring instead.

transmart-server/grails-app/services/org/transmart/server/subscription/QuerySetSubscriptionMailService.groovy
```diff
@@ -0,0 +1,139 @@
...
+     * - a name of the query,
+     * - list of added and removed ids of objects that the query relates to
+     * - over what period the change was
+     *
+     * @param username
+     * @param freq
+     * @return The body of the email
+     *
+     */
+    private String generateEmail(String username, SubscriptionFrequency frequency) {
```
forus: It's better to have small methods that do ideally one, or few things at once. You could have `private String generateEmail(List<UserQuerySetChangesRepresentation> querySetChanges)` instead.

transmart-server/grails-app/services/org/transmart/server/subscription/QuerySetSubscriptionMailService.groovy
```diff
@@ -0,0 +1,139 @@
...
+
+            StringBuilder textStringBuilder = new StringBuilder()
+            textStringBuilder.append("Generated as per day ${currentDate.format("d.' of 'MMMM Y h:mm aa z")}")
+            textStringBuilder.append(NEW_LINE)
+            textStringBuilder.append("List of updated query results:")
+            textStringBuilder.append(NEW_LINE)
+
+            def patientSets = querySetChanges.findAll { it.setType == SetType.PATIENT }?.sort { it.queryId }
+            if (patientSets.size() == 0) {
+                return null
```
forus: Try don't return `null`. In this particular example, there is no other type then patient set.

transmart-server/grails-app/jobs/org/transmart/server/subscription/QuerySetSubscriptionDailyJob.groovy
```diff
@@ -0,0 +1,46 @@
...
+         Day of Month, 1-31, ?;
+         Month, 1-12 or JAN-DEC;
+         Day of Week, 1-7 or SUN-SAT, ?;
+         Year [optional]
+      Either Day-of-Week or Day-of-Month must be "?".
+      Can't specify both fields, nor leave both as the all values wildcard "*"
+     **/
+    static triggers = {
+        cron name: 'dailySubscriptionTrigger', cronExpression: " 0 0 0 1/1 * ?"
+        // cron name: 'dailySubscriptionTestTrigger', cronExpression: "0 0/1 * 1/1 * ?"
```
forus: please remove commented code

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/UserQueryController.groovy
```diff
@@ -91,9 +112,27 @@ class UserQueryController {
...
                     bookmarked       : bookmarked,
+                    subscribed       : subscribed,
+                    subscriptionFreq : subscriptionFreq,
                     createDate       : createDate,
                     updateDate       : updateDate,
             ]
         }
     }
+
+    private static void validateQuery(String query) {
```
forus: It does not validate the query, but json.

transmart-core-db/grails-app/services/org/transmartproject/db/user/UsersResourceService.groovy
```diff
@@ -66,4 +66,11 @@ class UsersResourceService implements UsersResource {
         } as List<CoreUser>
     }
 
+    @Transactional(readOnly = true)
+    @Override
+    List<CoreUser> getUsersWithEmailSpecified() {
+        User.withSession { session ->
```
forus: the left join produces one user row per role. Thus you would get list of users with duplicates. Remove the left join or deduplicate users afterward.

transmart-core-api/src/main/groovy/org/transmartproject/core/userquery/SetType.groovy
```diff
@@ -0,0 +1,31 @@
...
+import groovy.transform.CompileStatic
+import groovy.util.logging.Slf4j
+
+/**
+ * Supported types of sets for user queries
+ */
+@CompileStatic
+@Slf4j
+enum SetType {
+    PATIENT ('PATIENT')
```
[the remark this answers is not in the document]
ewelinagr: As I heard sample sets will be introduced soon, so SetType will be extended. Is it for sure worth removing (SetType is used in multiple places)?

open-api/swagger.json
```diff
@@ -2263,6 +2263,12 @@
                 },
                 "bookmarked": {
                   "type": "boolean"
+                },
+                "subscribed": {
```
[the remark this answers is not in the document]
ewelinagr: I can also change the UserQueryController to prevent having a situation like this. So it will ensure that: if subscribed = false, subscriptionFreq has to be null.
Would it solve the issue?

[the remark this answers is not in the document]
forus: I am not sure whether it's even the issue. We could leave it like that.

transmart-core-api/src/main/groovy/org/transmartproject/core/userquery/SetType.groovy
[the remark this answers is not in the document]
forus: I don't think so then. Let's leave it like that.

transmart-core-db/grails-app/services/org/transmartproject/db/user/UsersResourceService.groovy
ewelinagr: thanks Ruslan, this is fixed now

transmart-server/grails-app/services/org/transmart/server/subscription/QuerySetSubscriptionMailService.groovy
```diff
@@ -0,0 +1,134 @@
+package org.transmart.server.subscription
```
gijskant: These files are now the only functional parts of `transmart-server`, while this module was meant for only bundling components. Please, move the subscription functionality to either a separate component (could be a grails plugin or a plain groovy library, but then you need to add some Spring annotations), or to, e.g., core-db. `transmart-notifications`, `transmart-email` or something.

forus:
Merged with dev branch instead.

PR #148, opened 2018-01-29
Add nullValueCounts to aggregatePerConcept results 

ewelinagr:
NTRDEV-279

### DIFF
--- a/open-api/swagger.json
+++ b/open-api/swagger.json
@@ -3176,6 +3176,10 @@
               "additionalProperties": {
                 "type": "number"
               }
+            },
+            "nullValueCounts": {
+              "description": "count of null values",
+              "type": "number"
             }
           }
         }
--- a/open-api/swagger.yaml
+++ b/open-api/swagger.yaml
@@ -2490,3 +2490,6 @@ definitions:
                 type: number
             additionalProperties:
               type: number
+          nullValueCounts:
+            description: count of null values
+            type: number
--- a/open-api/swagger_spec.js
+++ b/open-api/swagger_spec.js
@@ -3176,6 +3176,10 @@ var spec = {
               "additionalProperties": {
                 "type": "number"
               }
+            },
+            "nullValueCounts": {
+              "description": "count of null values",
+              "type": "number"
             }
           }
         }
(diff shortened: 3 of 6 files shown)


gijskant:
Apparently, a test is failing: https://travis-ci.org/thehyve/transmart-core/jobs/334743334#L3090.
Makes me wonder, with this patch we cannot distinguish between `null` values and empty strings... not sure if that's desired actually. If you fill in the empty string in a value constraint, you won't get the empty values back (null != '').

ewelinagr:
hmm true,
so the current format is:
`{
	"aggregatesPerConcept": {
		"gender": {
			"categoricalValueAggregates": {
				"valueCounts": {
					'': 1,
					"Male": 1
				}
			}
		}
	}
}`
the 'empty string' is a key of the map (not a value as I thought before). So to directly use null values it should be:
`{
	"aggregatesPerConcept": {
		"gender": {
			"categoricalValueAggregates": {
				"valueCounts": {
					null: 1,
					"Male": 1
				}
			}
		}
	}
}` 
but it is not a valid JSON then

gijskant:
Good point.
I'm also not sure if we can have empty strings as valid values. We could decide on empty (null or '') to be a missing/empty value. For numeric it is different, but there we aggregate in a different way anyway.
So then, the test should be fixed and we need to thing about how this should be reflected in the value constraints: should '' be interpreted as null there as well?

JanKanis:
I think we should distinguish null and "". Null is a valid object key in javascript, but many json libraries apparently don't like it. Maybe a different way to encode it?

gijskant:
Okay, apparently no consensus yet ;)

JanKanis:
My idea would be to put a key "nullKey": "somestring" somewhere in the json output and then use "somestring" instead of null. Select "somestring" in such a way that it is not a key in this json output. e.g. "null", if that exists already "null_", "null__", etc.

gijskant:
We could extend the class [CategoricalValueAggregates](https://github.com/thehyve/transmart-core/blob/master/transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/CategoricalValueAggregates.groovy) with an integer field `nullValues` instead.

PR #149, opened 2018-01-30
All changes happen in single transaction

forus:


transmart-copy/README.md
```diff
@@ -29,6 +29,7 @@ _Parameters:_
 - `-w <file>`, `--write <file>`: Write observations to TSV file `<file>`.
 - `-v`, `--vacuum-analyze`: Vacuum analyze the `observation_fact` table.
 - `-d`, `--directory`: Specifies a data directory.
+- `-ru`, `--re-upload`: Removes and loads the studies specified in the data directory.
```
gijskant: But why not reuse the `-d` parameter then?

transmart-copy/src/test/groovy/org/transmartproject/copy/CopySpec.groovy
```diff
@@ -1,24 +1,23 @@
...
-    static STUDY_FOLDER = './src/main/resources/examples/' + TEST_STUDY
-    static DATABASE_CREDENTIALS = [
-            PGUSER: 'i2b2demodata',
+    static String STUDY_FOLDER = './src/main/resources/examples/' + TEST_STUDY
+    static String CORRUPTED_STUDY_FOLDER = './src/main/resources/examples/' + TEST_STUDY + '_corrupted'
+    static Map<String, String> DATABASE_CREDENTIALS = [
+            PGUSER    : 'i2b2demodata',
             PGPASSWORD: 'i2b2demodata'
-    ]
+    ].withDefault { String key -> System.getenv(key) }
```
gijskant: Is there a reason to include the tester's environment variables in the test?

transmart-copy/src/main/groovy/org/transmartproject/copy/Copy.groovy
```diff
@@ -140,61 +127,90 @@ class Copy {
...
                 def studyId = cl.getOptionValue('delete')
-                try {
-                    copy.deleteStudy(studyId)
-                } catch(Exception e) {
-                    log.error "Error deleting study ${studyId}."
-                    throw e
-                }
+                copy.deleteStudyById(studyId)
+            }
+            if (cl.hasOption('re-upload')) {
```
gijskant: Twice checking for re-upload...

transmart-copy/src/main/groovy/org/transmartproject/copy/Copy.groovy
```diff
@@ -140,61 +127,90 @@ class Copy {
...
+                def directory = cl.getOptionValue('re-upload')
+                copy.deleteStudy(directory, false)
             }
             if (cl.hasOption('drop-indexes')) {
                 copy.dropIndexes()
             }
-            if (cl.hasOption('mode')) {
+            boolean independentOperationOptionSpecified = INDEPENDENT_OPERATION_OPTIONS
```
gijskant: The control flow is really unclear here.

transmart-copy/README.md
forus: ah, true. So rewriting a study would be the default behavior.

transmart-copy/src/main/groovy/org/transmartproject/copy/Copy.groovy
forus: It became a little bit simpler by removing `--re-upload` option ;)

PR #155, opened 2018-02-06
Parallel export

gijskant:
Main contributions in these changes:
- Parallel implementation of counts per study and concept;
- Parallel implementation of survey table export;
- Refactor of the MultiDimensionalDataResourceService (split in two);
- Canonise constraints;
- Enable reuse of existing patient sets by matching the constraint.


transmart-core-db/grails-app/services/org/transmartproject/db/clinical/SurveyTableColumnService.groovy
```diff
@@ -0,0 +1,298 @@
...
+     * Get a list of all unique study-concept pairs for a data set based on the constraint.
+     * The list is computed using a counts per study and concept query on the database.
+     * These counts may be already in the application cache.
+     * The query is restricted to the data the current user has access to.
+     *
+     * @param constraint the constraints that defines the data set.
+     * @param user the current user.
+     * @return the list of all unique study-concept pairs for the data set.
+     */
+    private List<StudyConceptPair> getStudyAndConceptListFromCounts(MultiDimConstraint constraint, User user) {
```
forus: `getStudyConceptPairs` instaed?

transmart-rest-api/grails-app/services/org/transmartproject/rest/SurveyTableViewDataSerializationService.groovy
```diff
@@ -32,31 +48,156 @@ import org.transmartproject.rest.serialization.tabular.TabularResultTSVSerialize
...
+     */
     @Override
     void writeClinical(Format format,
                        MultiDimConstraint constraint,
                        User user,
                        OutputStream out) {
+        log.info "Start parallel export ..."
+        PatientSetConstraint patientSetConstraint = null
+        Constraint columnConstraint = new TrueConstraint()
+        if (constraint instanceof PatientSetConstraint) {
```
forus: Could you try to get constraint optimization code in a separate class? It looks like it could be brought to the rest of such code in `SurveyTableColumnService` class at least. Also the constraint optimization code looks pretty massive. We should cover this code with some unit tests.

gijskant:
@forus I changed the implementation based on your feedback. There are two things that still need attention:
- the export fails for a hypercube without the StartDateDimension;
- check if the parallel code is covered by tests.

gijskant:
The export now also works if no start time dimension is available. There is an integration test for parallel counting, not yet for parallel export.

gijskant:
@forus Rest API tests fixed.

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -681,13 +418,18 @@ class MultidimensionalDataResourceService implements MultiDimensionalDataResourc
                 .insertResultToTable(QtPatientSetCollection, ['patient.id', 'resultInstance.id'], patientSetCriteria)
     }
 
-    @Override QueryResult findQueryResult(Long queryResultId, User user) {
-        QueryResult queryResult = QtQueryResultInstance.findById(queryResultId)
+    @Override
+    QueryResult findQueryResult(Long queryResultId, User user) {
```
forus: Now, actually, it returns something else. `findFinishedPatientSetQueryResult(...)`
It change contract of `GET /v2/patient_sets/${id}` call as this method is reused.

transmart-core-db/grails-app/services/org/transmartproject/db/support/ParallelPatientSetTaskService.groovy
```diff
@@ -0,0 +1,136 @@
...
+     * @param user
+     * @param sequentialTask the sequential fallback implementation.
+     * @param parallelSubTask the subtask implementation.
+     * @param combine the function that combines the results of the subtasks.
+     * @return the result of the combination of subtasks, or of the sequential fallback.
+     */
+    public <ResultType, SubtaskResultEntry> ResultType run(
+            TaskParameters parameters,
+            Function<TaskParameters, ResultType> sequentialTask,
+            Function<SubtaskParameters, List<SubtaskResultEntry>> parallelSubTask,
```
forus: Wouldn't `parallelSubTask` suffice?  `parallelSubTask` would be basically `sequentialTask` when number of tasks = `1`.

transmart-core-api/src/main/groovy/org/transmartproject/core/config/SystemResource.groovy
```diff
@@ -0,0 +1,15 @@
+package org.transmartproject.core.config
+
+interface SystemResource {
+
+    RuntimeConfig getRuntimeConfig()
+
+    RuntimeConfig updateRuntimeConfig(RuntimeConfig config)
```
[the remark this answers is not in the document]
gijskant: Now it is possible to ignore the result, but also to use it.

transmart-core-db/grails-app/services/org/transmartproject/db/support/ParallelPatientSetTaskService.groovy
gijskant: Agreed. I will change it.

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
gijskant: Which is intentional. Actually `GET /v2/patient_sets` should also only return finished patient sets.

transmart-core-api/src/main/groovy/org/transmartproject/core/config/SystemResource.groovy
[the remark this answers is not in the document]
forus: Ah, like that. Why would you want to ignore the config? Doesn't it sound like an exception then?

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
forus: Yes, actually, they are always finished currently. As the call is synchronous.

gijskant: Unless multiple concurrent requests of course.

gijskant: And then you only want the finished ones.

PR #162, opened 2018-02-13
Add support for null values in query builder

ewelinagr:


transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/query/HibernateCriteriaQueryBuilder.groovy
```diff
@@ -253,7 +253,7 @@ class HibernateCriteriaQueryBuilder implements QueryBuilder<Criterion, DetachedC
                     if (Number.isAssignableFrom(fieldType) && value instanceof Date) {
                         convertedValue = toNumber(value)
                     } else {
-                        convertedValue = fieldType.newInstance(value)
+                        convertedValue = value ? fieldType.newInstance(value) : null
```
gijskant: Please explicitly use `value == null` here, the value might be `''`, `0`, `false` or something else that groovy may interpret as `false`.

PR #168, opened 2018-02-19
Use Jackson for deserialising constraints. 

gijskant:


transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/query/Constraint.groovy
```diff
@@ -184,13 +195,14 @@ enum Operator {
...
  */
+@CompileStatic
 @Canonical
 @Sortable
 class Field {
     String dimension
     @NotNull
     Type type = Type.NONE
-    @NotBlank
+    @Size(min = 1)
```
forus: A string with one space would be considered a valid.

transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/query/Constraint.groovy
```diff
@@ -204,6 +216,31 @@ class Field {
...
+    @JsonSubTypes.Type(value = Combination.class, name = 'combination'),
+    @JsonSubTypes.Type(value = AndConstraint.class, name = 'and'),
+    @JsonSubTypes.Type(value = OrConstraint.class, name = 'or'),
+    @JsonSubTypes.Type(value = TemporalConstraint.class, name = 'temporal'),
+    @JsonSubTypes.Type(value = ConceptConstraint.class, name = 'concept'),
+    @JsonSubTypes.Type(value = StudyNameConstraint.class, name = 'study_name'),
+    @JsonSubTypes.Type(value = StudyObjectConstraint.class, name = 'study'),
+    @JsonSubTypes.Type(value = NullConstraint.class, name = 'null'),
+    @JsonSubTypes.Type(value = SubSelectionConstraint.class, name = 'subselection'),
+    @JsonSubTypes.Type(value = RelationConstraint.class, name = 'relation')
```
forus: what about `name = RelationConstraint.name`?

PR #176, opened 2018-02-22
Documentation enhancements

mickmis:
- add external links to other installation guides
- add versions of dependencies
- other small modifications

Addresses TMNP-36 and part of TMNP-38.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -64,7 +64,8 @@ tranSMART requires a configuration file to be generated in `~/.grails/transmartC
 Scripts to generate the configuration are shipped with
 `transmart-data`.
 
-Fetch `transmart-data`:
+Either use the sources in this repository (`transmart-core/transmart-data`),
+or fetch `transmart-data` from a Nexus repository:
 ```
 mvn dependency:get -Dartifact=org.transmartproject:transmart-data:17.1-SNAPSHOT:tar -DremoteRepositories=https://repo.thehyve.nl/content/repositories/snapshots/
 mvn dependency:unpack -Dartifact=org.transmartproject:transmart-data:17.1-SNAPSHOT:tar -DoutputDirectory=.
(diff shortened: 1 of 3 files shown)


transmart-data/README.md
```diff
@@ -76,6 +77,9 @@ The several options are fairly self-explanatory.
 
 ### PostgreSQL-specific notes
 
+These scripts were tested with PostgreSQL version 9.5. PostgreSQL version 10.2
```
forus: @mickmis  any details on what did not work?

transmart-data/README.md
```diff
@@ -45,7 +46,7 @@ The following are required:
   `PATH` (Oracle and some secondary functionality only)
 * [Kettle][kettle] (ETL only)
 * rsync (Solr only)
-* Java environment (JDK 7)
+* Java environment (JDK 7 or 8)
```
forus: Does it work with java 7?

docs/installation.md
```diff
@@ -104,9 +113,12 @@ mvn dependency:copy -Dartifact=org.transmartproject:transmart-server:17.1-SNAPSH
 java -jar transmart-server-17.1-SNAPSHOT.war
 ```
 
-## 6. Start services
 
-`Rserve` and `Solr` services can be run using `transmart-server` (to fetch `transmart-data`, see [Generate configuration files](#3-generate-configuration-files))
+## 5. Start services
+
+`Rserve` and `Solr` services can be run using `transmart-server` (to fetch `transmart-data`,
```
forus: Did you mean using `transmart-data`?

transmart-data/README.md
mickmis: At least one thing, which is the creation of the database. The error is:
```bash
/Library/PostgreSQL/10/bin/psql -v ON_ERROR_STOP=1 --single-transaction -X -d template1 -c "CREATE DATABASE transmart TEMPLATE template0 \
      ENCODING='UTF-8' LC_COLLATE='en_US.UTF-8'"
ERROR:  CREATE DATABASE cannot run inside a transaction block
```
Because this is the very beginning of the loading script, others things could be broken with PostgreSQL 10.2.

transmart-data/README.md
mickmis: I don't know, did not test. My understanding was that it works since it was there before.

docs/installation.md
mickmis: I didn't notice this typo since I actually skipped this part, will fix.

transmart-data/README.md
forus: well, for `transmart-data` java 7 would work, yes. I remember having problems to run the web app though.

transmart-data/README.md
forus: ah, sounds familiar. Its something to fix for the future releases.

PR #178, opened 2018-02-28
Feature/GitHub pages ewelina

forus:


gijskant:
@ewelinagr @forus @mickmis I see a dead link in the README.md to https://username.github.io/transmart-core/. Where is this hosted?

gijskant:
Ah, okay, and since the dev branch is made default now, that breaks the front page. We'll have to think about how to solve that.

PR #193, opened 2018-03-19
Add docs to Hypercube and HypercubeValue

JanKanis:


transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/HypercubeValue.groovy
```diff
@@ -2,21 +2,57 @@
...
+     * Indexes are unique to a dimension element only within one hypercube and within that dimension. Indexes are
+     * guaranteed to be contiguous, starting from 0.
+     *
      * @param dim The dimension
-     * @return The index of the element for this dimension in the list returned by Hypercube.dimensionElements(dim).
-     * If this value does not have an element for this index, returns null.
+     * @return The index of the dimension element for the specified dimension. If this HypercubeValue does not have
+     * an element for this dimension, returns null.
+     * @throws UnsupportedOperationException if dim is not a {@link Dimension.Density#DENSE} dimension
+     * @throws IllegalArgumentException if dim is not part of this Hypercube
```
ewelinagr: this one is thrown also by other methods, so could be added also to their description, but it's just a small thing

PR #194, opened 2018-03-19
Add queryBlob to user queries

ewelinagr:
To support storing of a data table state in the UI
TMT-85, TMT-86


JanKanis:
Looks fine. I would have called it 'userData' instead of 'queryBlob', but it doesn't matter a lot.

PR #195, opened 2018-03-20
Sorting

JanKanis:
Support sorting in the hypercube and in the /v2/observations call

### DIFF
--- a/open-api/swagger.yaml
+++ b/open-api/swagger.yaml
@@ -753,6 +753,25 @@ paths:
           description: 'json that specifies the constraint. Example: `{"type":"study_name", "studyId":"EHR"}` or
           `{"type":"concept","path":"\\Public Studies\\CLINICAL_TRIAL_HIGHDIM\\High Dimensional data\\Expression Lung\\"}`.'
           type: string
+        - name: sort
+          required: false
+          in: query
+          type: string
+          description: >
+              json string that specifies the sort order of the observations. Sorting can be done on multiple
+              dimensions. `["patient"]` sorts the observations on the patient dimension, `["patient", "concept"]`
+              sorts the observations first on the patient, and then on concept. The sort order is on the 'key'
+              field of the chosen dimension, sorting on arbitrary fields is not yet supported. For patients, this is
+              the `id`, for concepts the `conceptCode`.
+
+              It is also possible to specify the sort order to be ascending or descending. Use `[['patient', 'asc'],
+              ['concept', 'desc']]` for that to sort the observations on patients first, ascending, and then on
+              concepts descending.
+
+              When requesting modifier dimensions, the supported sortings is very limited due to implementation
+              constraints. Sorting support is limited to those dimensions that make up the primary key columns in the
+              `i2b2demodata.observation_fact` table, and a few other supported dimensions. If you request a sort
+              order that is not supported you will receive an HTTP 400 Bad Request error code.
         - name: biomarker_constraint
           required: false
           in: query
@@ -799,6 +818,9 @@ paths:
                     type: string
                     description: see GET parameters
                   projection:
+                    type: string
+                    description: see GET parameters
+                  sort:
                     type: string
                     description: see GET parameters
                 required: [ "type", "constraint" ]
--- a/transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
+++ b/transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
@@ -28,8 +28,8 @@ import org.transmartproject.core.dataquery.highdim.dataconstraints.DataConstrain
 import org.transmartproject.core.dataquery.highdim.projections.Projection
 import org.transmartproject.core.exceptions.EmptySetException
 import org.transmartproject.core.exceptions.InvalidArgumentsException
-import org.transmartproject.core.exceptions.InvalidRequestException
 import org.transmartproject.core.exceptions.NoSuchResourceException
+import org.transmartproject.core.exceptions.UnsupportedByDataTypeException
 import org.transmartproject.core.multidimquery.query.BiomarkerConstraint
 import org.transmartproject.core.multidimquery.query.Combination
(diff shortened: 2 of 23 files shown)


open-api/swagger.yaml
```diff
@@ -753,6 +753,25 @@ paths:
           description: 'json that specifies the constraint. Example: `{"type":"study_name", "studyId":"EHR"}` or
           `{"type":"concept","path":"\\Public Studies\\CLINICAL_TRIAL_HIGHDIM\\High Dimensional data\\Expression Lung\\"}`.'
           type: string
+        - name: sort
+          required: false
+          in: query
+          type: string
+          description: >
```
ewelinagr: Did you mean: `description: |` ? 
The '>' symbol is breaking a formatting here

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -89,9 +92,9 @@ class MultidimensionalDataResourceService extends AbstractDataResourceService im
      * @param constraints: (nullable) A list of Constraint-s. If null, selects all the data in the database.
      * @param dimensions: (nullable) A list of Dimension-s to select. Only dimensions valid for the selected studies
      * will actually be applied. If null, select all available dimensions.
-     *
-     * Not yet implemented:
-     * @param sort
+     * @param sort: (nullable) Either a list of dimensions, or a LinkedHashMap of dimensions to "asc" or "desc".
+     * Dimensions can be either dimension objects or their string names. Note: allowed sortings are limited if
+     * modifier dimensions are used.
```
ewelinagr: This description should be updated also in the core-api interface: [MultiDimensionalDataResource](https://github.com/thehyve/transmart-core/blob/sorting2/transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/MultiDimensionalDataResource.groovy#L20)

open-api/swagger.yaml
```diff
@@ -753,6 +753,25 @@ paths:
...
+              json string that specifies the sort order of the observations. Sorting can be done on multiple
+              dimensions. `["patient"]` sorts the observations on the patient dimension, `["patient", "concept"]`
+              sorts the observations first on the patient, and then on concept. The sort order is on the 'key'
+              field of the chosen dimension, sorting on arbitrary fields is not yet supported. For patients, this is
+              the `id`, for concepts the `conceptCode`.
+
+              It is also possible to specify the sort order to be ascending or descending. Use `[['patient', 'asc'],
+              ['concept', 'desc']]` for that to sort the observations on patients first, ascending, and then on
+              concepts descending.
```
ewelinagr: it may be a bit confusing that for multiple dimensions the args format is different, so it's a list of lists, not [[PATIENT: 'desc'], [CONCEPT: 'asc']], but this shouldn't be a big problem. Maybe it is useful to add an example of the order specification for a single dimension, since it's different?

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -108,17 +111,33 @@ class MultidimensionalDataResourceService extends AbstractDataResourceService im
...
+
+    private static Map parseSort(sort) {
+        if (sort == null) {
+            [:]
+        } else if (sort instanceof Map) {
+            sort.collectEntries { [toDimensionImpl(it.key), toSortOrder(it.value)] }
+        } else if (sort instanceof List) {
+            sort.collectEntries {
+                it instanceof List ?
+                        [toDimensionImpl(it[0]), toSortOrder(it[1])] :
```
ewelinagr: Couldn't this be Map instead of List (as for single dimension)?

open-api/swagger.yaml
JanKanis: What do you mean? This is valid yaml syntax, see e.g. https://yaml-multiline.info/. `|` preserves line breaks, `>` treats text as a wrapped line

open-api/swagger.yaml
JanKanis: True, the problem is that json doesn't have ordered maps, and this needs to be ordered. There are examples of the simpler syntax in the first paragraph.

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
JanKanis: You mean `[toDimensionImpl(it[0]): toSortOrder(it[1])]`? That also works, but it makes no difference as these two item lists or single item maps are collected by the `collectEntries` into a map.

`parseSort` accepts null, a map of dimension -> order, or a list where each item is either a dimension or a [dimension, order] pair (as a two-item list, to be able to parse the json representation). A dimension is either a string or a dimension object, a sort order is either a string or a SortOrder enum member. But I'll add some documentation to this method.

open-api/swagger.yaml
ewelinagr: I am not sure if swagger interprets it correctly, the formatting of the generated documentation looks strange starting from this symbol (comparing to other parameters before this): 
selection_073

JanKanis: Hmm that's strange. But I thought swagger used the swagger.json or swagger_spec.js files. There the syntax is just a normal javascript string.

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
JanKanis: I'll update that

open-api/swagger.yaml
JanKanis: The difference seems to be due to required vs optional parameters. The `type` and `constraint` parameters have an extra `required` css class. So looks like it works as intended.

ewelinagr: ah, you are right! sorry!

PR #196, opened 2018-03-21
Sorting part 2

JanKanis:
Encode the sorting used in the serialized hypercubes (json and protobuf).

This PR should go on top of the first sorting PR.

### DIFF
--- a/open-api/swagger.yaml
+++ b/open-api/swagger.yaml
@@ -2443,6 +2443,24 @@ definitions:
             type: array
             items:
               $ref: '#/definitions/dimensionDeclaration'
+          sort:
+            type: object
+            properties:
+              dimensionIndex:
+                type: number
+                description: >
+                    An index into the dimensionDeclarations list, indicating the dimension on which the
+                    result is sorted.
+              sortOrder:
+                type: string
+                enum: [asc, desc]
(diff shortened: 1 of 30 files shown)


open-api/swagger.yaml
```diff
@@ -2443,6 +2443,24 @@ definitions:
             type: array
             items:
               $ref: '#/definitions/dimensionDeclaration'
+          sort:
+            type: object
+            properties:
+              dimensionIndex:
+                type: number
+                description: >
```
ewelinagr: Should be `description: |`

open-api/swagger.yaml
```diff
@@ -2443,6 +2443,24 @@ definitions:
...
+                    An index into the dimensionDeclarations list, indicating the dimension on which the
+                    result is sorted.
+              sortOrder:
+                type: string
+                enum: [asc, desc]
```
ewelinagr: I am not sure if it's supported like this. Shouldn't 'type' and 'enum' be under 'schema' like here https://swagger.io/docs/specification/data-models/enums/ ?

JanKanis: Reading the documentation, I'm not sure either. It only gives an example for parameters, not model properties. But swagger treats it the same if it is under a `schema:` or not.

PR #200, opened 2018-03-28
 Add implementation of 'study' dimension elements selection

ewelinagr:
TMT-121

### DIFF
--- a/transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/DimensionImpl.groovy
+++ b/transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/DimensionImpl.groovy
@@ -623,13 +623,21 @@ class StudyDimension extends I2b2Dimension<MDStudy, Long> implements CompositeEl
     @Override List<MDStudy> doResolveElements(List<Long> elementKeys) {
         resolveWithInQuery(I2B2Study.createCriteria(), elementKeys)
     }
-
+    @Override
     DetachedCriteria selectDimensionElements(DetachedCriteria criteria) {
-        throw new InvalidArgumentsException("Retrieving elements of the $name dimension is not implemented.")
+        criteria.add(HibernateCriteriaQueryBuilder.defaultModifierCriterion)
+        criteria.setProjection(Projections.property('trialVisit'))
+
+        def dimensionCriteria = DetachedCriteria.forClass(I2B2Study, 'study')
+        dimensionCriteria.createAlias('trialVisits', 'trialVisits')
+        dimensionCriteria.add(Subqueries.propertyIn('trialVisits.id', criteria))
+        dimensionCriteria
+
     }
+
     @Override
     DetachedCriteria elementCount(DetachedCriteria criteria) {
-        throw new InvalidArgumentsException("Retrieving the element count of the $name dimension is not implemented.")
+        selectDimensionElements(criteria).setProjection(Projections.rowCount())
     }
 
 }
--- a/open-api/swagger.json
+++ b/open-api/swagger.json
@@ -568,7 +568,7 @@
         }
       }
     },
-    "/v2/studyId/{studyId}": {
+    "/v2/studies/studyId/{studyId}": {
       "get": {
         "description": "Gets the study with study id `id`.\n",
         "tags": [
--- a/open-api/swagger.yaml
+++ b/open-api/swagger.yaml
@@ -688,7 +688,7 @@ paths:
         404:
           description: |
             Study not found
-  /v2/studyId/{studyId}:
+  /v2/studies/studyId/{studyId}:
     get:
       description: |
         Gets the study with study id `id`.
--- a/open-api/swagger_spec.js
+++ b/open-api/swagger_spec.js
@@ -568,7 +568,7 @@ var spec = {
         }
       }
     },
-    "/v2/studyId/{studyId}": {
+    "/v2/studies/studyId/{studyId}": {
       "get": {
         "description": "Gets the study with study id `id`.\n",
         "tags": [
--- a/transmart-core-db-tests/src/integration-test/groovy/org/transmartproject/db/multidimquery/QueryServiceSpec.groovy
+++ b/transmart-core-db-tests/src/integration-test/groovy/org/transmartproject/db/multidimquery/QueryServiceSpec.groovy
@@ -515,6 +515,49 @@ class QueryServiceSpec extends TransmartSpecification {
         locationsCount == Long.valueOf(expectedResults.grep().size())
     }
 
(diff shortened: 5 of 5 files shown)


gijskant:
Why replacing the endpoint instead of adding a new `"/studies/studyIds/$studyIds"`?

ewelinagr:
@gijskant I had some doubts about it. `"/studies/studyId/$studyId"`was partially a duplicate of "/studies/studyIds/$studyIds" since the results of e.g. `"/studies/studyId/study1"` and `"/studies/studyIds/study1"` would be the same.

JanKanis:
Having the call look like `/studies/studyIds/22,33,44/` looks a bit strange. Maybe it is better to make it a GET parameter encoding a json list. And keep the existing endpoint for a single study.

transmart-core-db/grails-app/services/org/transmartproject/db/ontology/MDStudiesService.groovy
```diff
@@ -107,17 +107,22 @@ class MDStudiesService implements MDStudiesResource, ApplicationRunner {
...
         def user = usersResource.getUserFromUsername(currentUser.username)
-        def study = Study.findByStudyId(studyId)
-        if (isLegacyStudy(study)) {
-            study = null
-        }
-        if (study == null || !user.canPerform(ProtectedOperation.WellKnownOperations.READ, study)) {
-            throw new AccessDeniedException("Access denied to study or study does not exist: ${studyId}")
+        def studies = []
+        for (studyId in studyIds) {
+            def study = Study.findByStudyId(studyId)
```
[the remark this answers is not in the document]
ewelinagr: I added one-by-one queries to have additional checks per study - not to return legacy studies and to throw "Access denied to study or study does not exist: " error, but if these checks are not needed (as you mentioned below), I will change it to one query for all studyIds

transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/DimensionImpl.groovy
```diff
@@ -623,13 +623,21 @@ class StudyDimension extends I2b2Dimension<MDStudy, Long> implements CompositeEl
...
-        throw new InvalidArgumentsException("Retrieving elements of the $name dimension is not implemented.")
+        criteria.add(HibernateCriteriaQueryBuilder.defaultModifierCriterion)
+        criteria.setProjection(Projections.property('trialVisit'))
+
+        def dimensionCriteria = DetachedCriteria.forClass(I2B2Study, 'study')
+        dimensionCriteria.createAlias('trialVisits', 'trialVisits')
+        dimensionCriteria.add(Subqueries.propertyIn('trialVisits.id', criteria))
+        dimensionCriteria
+
     }
```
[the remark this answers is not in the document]
ewelinagr: Not directly in this PR, but this is needed for the same functionality on front-end side - retrieving available dimensions based on specified constraints. So first it uses this implementation to get available studyIds based on constraints and then this list of studyIds is passed as parameter to the study call.

ewelinagr:
@JanKanis could you verify again?

transmart-core-db/grails-app/services/org/transmartproject/db/ontology/MDStudiesService.groovy
```diff
@@ -106,23 +108,34 @@ class MDStudiesService implements MDStudiesResource, ApplicationRunner {
...
+        def study = Study.findByStudyId(studyId)
+        if (isLegacyStudy(study)) {
+            study = null
+        }
+        if (study == null || !user.canPerform(ProtectedOperation.WellKnownOperations.READ, study)) {
+            throw new AccessDeniedException("Access denied to study or study does not exist: ${studyId}")
+        }
+        study.dimensions.size()
+        study
+    }
```
JanKanis: This looks like unnecessary code duplication. Either change the clients to use `getStudyByStudyIdsForUser([studyId], user)[0]` or change the implementation of this to delegate to the new method in the same way.

By the way, this is an interesting case study for you code quality goal. You might want to consider if/why you didn't catch this yourself and how you might prevent that in the future.

JanKanis: Whether to change the implementation of `getStudyByStudyIdForUser` or to change the clients is a design decision with (imo) not a clear answer. It's a trade off between simplicity in the MDStudiesService interface and simplicity in the clients. Though if there are more than two clients (or expected to be soon), I would opt for keeping the interface method and delegating the implementation.

ewelinagr: These lines are not actually a part of my PR, I just restored a method that I removed in my previous commit (getStudyByStudyIdForUser), so this duplication is already part of the main code. But I agree, I should have noticed that and improve it. I will add a change here.

JanKanis: According to github, this PR introduces the `getStudiesByStudyIdsForUser` call, thus introducing the duplication :). But I can understand you didn't notice it over the separate commits.

ewelinagr: the duplication is mainly between `getStudyForUser` and `getStudyByStudyIdForUser`, in `getStudiesByStudyIdsForUser` the implementation is a bit different, because it does not throw an error, when a user has no access or for legacy studies (just does not include these studies in the results). Should this be changed?

JanKanis:
I've added some changes. Some more code quality items to think about:

- The check in `checkAccessToStudy` was still duplicated. Removing that introduces some boilerplate to deal with the exceptions, but with such access checks (and business logic in general) it is easy to later modify one of them and not update the other copy, unlike non-business logic which would likely cause an error somewhere else. So for business logic and especially security checks it is good to have the logic in only one place

- The `usersResource.getUserFromUsername(currentUser.username)` in `checkAccessToStudy` is not necessary as far as I can see, it just loads the User object again by its username. You didn't introduce this line but if you are refactoring a piece of code you should consider why some code is needed, especially if it looks like it does nothing. (In this case, AFAICT, it is only needed in the `getStudies` call that uses `getdimensionStudiesForUser` that uses methods of the DbUser object that are not part of the User interface and not implemented by all User implementations)

- I changed all the calls to throw NoSuchResourceExceptions instead of AccessDeniedExceptions to avoid information leaks. When I first mentioned that in the previous review I wasn't aware that all these other methods were also returning access denied errors. But reviewers are not always right, so you could have mentioned the inconsistency.

JanKanis:
I'll merge when the tests finish

PR #201, opened 2018-03-29
Tabular result remove closures

JanKanis:
Refactor TabularResult implementations to use @CompileStatic and not use closures


transmart-core-db/src/main/groovy/org/transmartproject/db/dataquery/highdim/snp_lz/SnpLzModule.groovy
```diff
@@ -291,7 +291,7 @@ class SnpLzModule extends AbstractHighDimensionDataTypeModule {
         criteriaBuilder
     }
 
-    @Override
+    @Override //@CompileStatic
```
ewelinagr: why is @CompileStatic commented out here?

JanKanis: That method should be compilestatic for consistency with other such methods. But in this specific case the method doesn't use any of the classes that were modified, and the returned TabularResults are used as-is and not extended inline. So the benefit is limited in this case so I didn't refactor this method.

PR #208, opened 2018-04-10
Fix protobuf tests

JanKanis:
Fix tests that were failing by supporting serializing of maps/objects to protobuf messages. 

All tests should pass again after this.


ewelinagr:
Looks good to me, so just the failing test and merge conflicts

PR #212, opened 2018-04-16
Change the query master constraint index to use hash, not btree.

gijskant:


gijskant:
@forus This one is okay to merge?

forus:
@gijskant yes, it's fine.

PR #218, opened 2018-04-23
Add storage of a study specific table representation

ewelinagr:
TMT-95


transmart-core-api/src/main/groovy/org/transmartproject/core/ontology/StudyMetadata.groovy
```diff
@@ -12,6 +12,7 @@ class StudyMetadata {
     private static final JsonSlurper JSON_SLURPER = new JsonSlurper()
 
     Map<String, VariableMetadata> conceptCodeToVariableMetadata
+    TabularRepresentationMetadata defaultTabularRepresentation
 
     static StudyMetadata fromJson(String jsonText) {
```
gijskant: Looks like it's time to use ObjectMapper, see, e.g., [ConstraintFactory](https://github.com/thehyve/transmart-core/blob/dev/transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/query/ConstraintFactory.groovy#L59).

PR #219, opened 2018-04-23
Replace studyIds params in /v2/studies/studyIds call with a JSON encoded list

ewelinagr:
Replace `/v2/studies/studyIds?studyIds=foo&studyIds=bar ` with `/v2/studies/studyIds?studyIds=["foo","bar"] `

### DIFF
--- a/transmart-rest-api/grails-app/controllers/org/transmartproject/rest/StudyQueryController.groovy
+++ b/transmart-rest-api/grails-app/controllers/org/transmartproject/rest/StudyQueryController.groovy
@@ -13,6 +13,7 @@ import org.transmartproject.rest.marshallers.StudyWrapper
 
 import static java.util.Objects.requireNonNull
 import static org.transmartproject.rest.misc.RequestUtils.checkForUnsupportedParams
+import static org.transmartproject.rest.misc.RequestUtils.parseJson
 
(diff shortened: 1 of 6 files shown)


transmart-rest-api/grails-app/controllers/org/transmartproject/rest/StudyQueryController.groovy
```diff
@@ -87,22 +88,25 @@ class StudyQueryController extends AbstractQueryController {
...
      *
-     * @param a list of the study names
+     * @param a json-encoded list of the study ids ${studyIds}
      *
-     * @return a list of the {@link org.transmartproject.db.i2b2data.Study} objects with names ${studyIds}
+     * @return a list of the {@link org.transmartproject.db.i2b2data.Study}
      * if all of them exist and the user has access; null otherwise.
      */
     def findStudiesByStudyIds(@RequestParam('api_version') String apiVersion) {
         if (params.studyIds == null) {
```
gijskant: Is it not possible to just specify `List<String> studyIds` as parameter of `findStudiesByStudyIds`?

gijskant: Never mind, Grails is a bit limited here (only path variable and command objects for the body seem to work).

PR #221, opened 2018-04-24
Datatable export

JanKanis:
- Split DataTableImpl into a PagingDataTableImpl and a FullDataTable and an AbstractDataTable, and also split the interfaces similarly
- Implementation for TSV export of the data table
- tests, docs, and the regular stuff
- Fix tabular result DataSerializers to not secretly require a ZipOutputStream while the interface says they only need a regular OutputStream
- Add some magic to make tests work. Ingredients: a TestData.clearAllDataInTransaction() call to only clear data within the current transaction, and inheriting the tests from an empty base class. No idea why that helps. But it also worked for ObservationsBuilderTests which I moved to the right folder.


transmart-rest-api/src/main/groovy/org/transmartproject/rest/serialization/tabular/TabularResultSerializer.groovy
```diff
@@ -19,6 +19,10 @@
 package org.transmartproject.rest.serialization.tabular
 
 import org.transmartproject.core.dataquery.TabularResult
+import org.transmartproject.core.multidimquery.StreamingDataTable
+import org.transmartproject.core.users.User
+
+import java.util.zip.ZipOutputStream
```
ewelinagr: I guess this imports are not needed any more?

JanKanis:
Info on the failing tests:

I ran into two problems with the tests: the first caused the data table tests to fail with an exception that it could not load ApplicationContext, so something with the grails/spring setup not going right. That was fixed by deriving the test class from an empty intermediate class and applying the right annotations. Based on the errors from travis, this is working correctly there as well. Note that these errors may be fixed by a 'grails test clean' command on your local machine.

The second problem caused *other* tests to fail, because the existing rest-api integration tests don't set up their own data before each test, but instead rely on the setup at the beginning of the test run and each test running in a transaction that is rolled back at the end. I fixed this by creating a variant of TestData.clearAllData() that should only work within the current transaction. As the data table tests don't test the rest api endpoints, all db access done there happens within the same transaction. On my local machine this fixed it, but on Travis it doesn't. This issue is manifested by all tests following the data table tests failing because transmart says the study they need does not exist. 
As a workaround the last commits remove some of these changes, and don't clear any data at all at the start of the data table tests. (That is fine for the data table tests, which don't mind other data being present.) That fixes most of the other tests, but there are still 5 tests breaking from the v2 data that is loaded by the data table tests. 

So it looks like for some reason the tests are not being run in their own transaction in Travis, or the transaction is not being rolled back properly or something. To me this sounds like it might be related to using different versions of H2, grails, or some other dependency in Travis. 

A workaround might be to just fix the 5 failing tests, but that doesn't solve the underlying problem of course.

ewelinagr:
Looks good to me now!

PR #223, opened 2018-05-03
 Add a call for updates required after data loading

ewelinagr:
TMT-163
- clear patient sets
- clear patient bitset
- clear caches
- scan for user subscribed queries updates

### DIFF
--- a/open-api/swagger.json
+++ b/open-api/swagger.json
@@ -1566,7 +1566,7 @@
         }
       }
     },
-    "/v2/tree_nodes/clear_and_update": {
+    "/v2/system/after_data_loading_update": {
       "get": {
         "description": "This endpoint should be called after loading, deleting or updating data in the database. Clears tree nodes, counts caches, patient sets and bitsets. Updates data for subscribed user queries.\nOnly for administrators.\n",
         "tags": [
--- a/open-api/swagger.json
+++ b/open-api/swagger.json
@@ -1566,6 +1566,19 @@
         }
       }
     },
+    "/v2/tree_nodes/clear_and_update": {
+      "get": {
+        "description": "This endpoint should be called after loading, deleting or updating data in the database. Clears tree nodes, counts caches, patient sets and bitsets. Updates data for subscribed user queries.\nOnly for administrators.\n",
+        "tags": [
+          "v2"
+        ],
+        "responses": {
+          "200": {
+            "description": "The cache and patient sets have been cleared. The data for subscribed user queries is updated.\n"
+          }
+        }
+      }
+    },
     "/v2/storage": {
       "get": {
         "description": "Gets a list of all storage systems.\n",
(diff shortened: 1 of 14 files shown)


open-api/swagger.json
```diff
@@ -1566,6 +1566,19 @@
...
+        "tags": [
+          "v2"
+        ],
+        "responses": {
+          "200": {
+            "description": "The cache and patient sets have been cleared. The data for subscribed user queries is updated.\n"
+          }
+        }
+      }
+    },
```
ewelinagr: This call is an extension for "/v2/tree_nodes/clear_cache", but it is not a best fit for the /v2/tree_nodes call. @gijskant @forus do you have some suggestions on how to fit it in the API?

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/AggregateDataOptimisationsService.groovy
```diff
@@ -44,6 +44,13 @@ class AggregateDataOptimisationsService {
         dbViewExists('biomart_user', 'study_concept_patient_set_bitset')
     }
 
+    /**
+     * Refreshes biomart_user.study_concept_bitset
+     */
+    void clearPatientSetBitset() {
+        refreshMaterializedView('biomart_user', 'study_concept_bitset')
```
ewelinagr: @forus Do you have some ideas on what kind of test can be added for this part?

open-api/swagger.json
gijskant: You could create a separate controller, maybe a SystemController or MaintainanceController and have something like a /v2/system/clear_cache, /v2/maintainaince/dataloadingFinished, something descriptive and concise...

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -509,6 +509,16 @@ class MultidimensionalDataResourceService extends AbstractDataResourceService im
         )
     }
 
+    @Override
+    @Transactional
+    void clearPatientSets() {
+        log.info 'Clearing patient sets ...'
+        QtPatientSetCollection.executeUpdate('delete QtPatientSetCollection')
```
gijskant: An alternative implementation:
```sql
update i2b2demodata.qt_query_result_instance set delete_flag = 'Y';
truncate i2b2demodata.qt_patient_set_collection;
```

ewelinagr: I tried truncating: 
```
def session = (StatelessSessionImpl) sessionFactory.openStatelessSession()
session.createSQLQuery('truncate table qt_patient_set_collection').executeUpdate()
```
but there were some issues with the table access.
regarding other tables than QtPatientSetCollection, is it better to set delete_flag than clear them?

gijskant: We could always change it to truncate if that is needed for performance.
Setting the delete_flag instead of actually deleting is a different way to achieve the same result, but it requires less changes.
The QtQueryMaster records do not have to be removed, they are reused in successive queries.

gijskant: So, doing the delete on QtQueryInstance, QtQueryResultInstance and QtPatientSetCollection should be sufficient.

PR #228, opened 2018-05-14
Cross table

ewelinagr:
Create `/v2/observations/crosstable` endpoint that takes row and column constraints as input and produces a table of subject counts for these constraints.

TMT-178


transmart-core-db/grails-app/services/org/transmartproject/db/clinical/CrossTableService.groovy
```diff
@@ -0,0 +1,66 @@
...
+        for(Constraint constraint in constraints) {
+            try {
+                checkAccess(constraint, user)
+            } catch(AccessDeniedException e) {
+                log.warn e.message
+                return 0
+            }
+        }
+        Constraint crossConstraint = new Combination(Operator.AND, constraints)
+        return aggregateDataService.counts(crossConstraint, user).patientCount
```
ewelinagr: This method does most of the job to create the cross table, but I am not sure if that is all that is required for the cross table in terms of creating/reusing/cashing patient sets/bitsets. @gijskant is this enough?

gijskant: For using bitsets and reusing patient sets this is not enough, but that doesn't matter for now. This can be our first (reference) implementation that we can later optimise for large datasets easily, but saving a patient set for each of the (row and column) constraints.
I do think that instead of a combination it should be a multiplesubselect. This works fine as long as each of the row and column constraints are patient set constraints or subselects, but if they are value or concept constraints you will get empty results.

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/CrossTableService.groovy
```diff
@@ -0,0 +1,66 @@
...
+    PatientSetResource patientSetResource
+
+    @Autowired
+    AggregateDataService aggregateDataService
+
+    @Override
+    CrossTable retrieveCrossTable(List<Constraint> rowConstraints, List<Constraint> columnConstraints,
+                                           Long patientSetId, User user) {
+
+        Constraint patientSetConstraint = new PatientSetConstraint(patientSetId)
```
gijskant: Nice, this suggests that we could easily switch to passing arbitrary constraints instead of a patient set id. Which solution do you think is better? I initially thought passing a patient set id was a good idea, but it makes the cross table functionality less flexible.

ewelinagr: I also saw that supporting only patient set id may cause some problems (described in TMT-219). Would this be a good solution to support both or would it be too messy? Between id and  constraint I would rather change it to constraint. I agree it would be more flexible (it could be combined with `GET /v2/patient_sets/{resultInstanceId}` to get patient set constraint based on id) and would solve the mentioned issues.

transmart-core-db/grails-app/services/org/transmartproject/db/clinical/CrossTableService.groovy
ewelinagr: @gijskant @forus I wanted to replace Combination constraint with MultipleSubSelectionsConstraint (as suggested by Gijs), but [this line of AggregateDataService|https://github.com/thehyve/transmart-core/blob/dev/transmart-core-db/grails-app/services/org/transmartproject/db/clinical/AggregateDataService.groovy#L206] gives me UnsupportedOperationException pointing to ConstraintSerialiser.build(ConstraintSerialiser.groovy:129). MultipleSubSelectionsConstraint was changed to be unsupported in ConstraintSerialiser in [b673b06e|https://github.com/thehyve/transmart-core/commit/b673b06e02bacb59d3a530375a40e20a1bdbc8dc] because it should be for internal use only, but in this case I want to use it internally. Should I change AggregateDataService or ConstraintSerialiser then?

gijskant: Or use `new AndConstraint([rowConstraint, columnConstraint, subjectConstraint].collect { new SubSelection(it) })])` or something similar. It is equivalent to the multiplesubselections.

gijskant: So, wrapping each of the constraints in a subselect is sufficient. Multiple subselections is just a more compact way of writing that, with the limitation that you cannot serialise it.

ewelinagr: ah ok, thanks! I will replace MultipleSubSelectionsConstraint with a mix of And/SubSelection constraints then

PR #230, opened 2018-05-17
Change naming of exported SPSS files and directory

ewelinagr:
TMT-213

### DIFF
--- a/transmart-rest-api/src/main/groovy/org/transmartproject/rest/dataExport/ExportJobExecutor.groovy
+++ b/transmart-rest-api/src/main/groovy/org/transmartproject/rest/dataExport/ExportJobExecutor.groovy
@@ -9,6 +9,7 @@ import org.quartz.UnableToInterruptJobException
 import org.springframework.context.ApplicationContext
 import org.transmartproject.core.users.User
 
+import java.text.SimpleDateFormat
 import java.util.zip.ZipOutputStream
 
 /**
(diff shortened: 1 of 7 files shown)


transmart-rest-api/src/main/groovy/org/transmartproject/rest/dataExport/ExportJobExecutor.groovy
```diff
@@ -44,10 +45,10 @@ class ExportJobExecutor implements InterruptableJob, AutoCloseable {
         Long jobId = jobDataMap.jobId
         String jobName = jobDataMap.jobName
         User user = jobDataMap.user
-        String file = getFilePath(user, "${jobName}.zip")
+        String file = getFilePath(user, jobName)
```
forus: Does it mean the following requirement taken from the ticket:
`Change the naming convention for exports to include the time it was created (or last updated) e.g.: export-for-wim_2018-04-15_1246.zip`

transmart-rest-api/src/test/groovy/org/transmartproject/rest/serialization/tabular/TabularResultSPSSSerializerSpec.groovy
```diff
@@ -42,7 +42,7 @@ class TabularResultSPSSSerializerSpec extends Specification {
         def columns = ImmutableList.copyOf([] as List<DataColumn>)
 
         when: 'producing spss files for empty table'
-        new TabularResultSPSSSerializer(Mock(User), Mock(ZipOutputStream), columns)
+        new TabularResultSPSSSerializer(Mock(User), Mock(ZipOutputStream), columns, 'testFile')
```
forus: Please add a check for the file (zip, dir. and file) names to the tests.

ewelinagr: I added a check like this to the [export e2e tests](https://github.com/thehyve/transmart-core/blob/export-file-naming/transmart-rest-api-e2e/src/test/groovy/tests/rest/v2/DataExportSpec.groovy), because this test is meant to fail before the files are created

forus: Looks fine.
Does it satisfy the TMT-213 description? What about timestamp as a filename?

PR #237, opened 2018-05-28
TranSMART docker support

mickmis:
Support to run the tranSMART stack with Docker. 

* `transmart-server`: WildFly-based container image to run the tranSMART server code
* `transmart-database`: PostgreSQL-based container image to host the database containing test data
* `solr`: SOLR container image (adapted from an existing version)
* `docker-compose.xxx.yml`: docker-compose profiles to configure and run the stack


gijskant:
We took a different, more lightweight approach, which is available here: https://github.com/thehyve/transmart-core/tree/dev/docker.
This applies Liquibase scripts for database schema creation and updates at application startup. Not all schemas and tables are supported, but the most relevant are.
See also [glowing-bear-docker](https://github.com/thehyve/glowing-bear-docker) for a more extensive setup, including the Glowing Bear user interface, TranSMART Packer export tool and Glowing Bear backend.

PR #238, opened 2018-05-31
 Replace core db user with core api one.

ewelinagr:


transmart-core-api/src/main/groovy/org/transmartproject/core/users/User.groovy
```diff
@@ -1,20 +1,27 @@
...
+
 /**
  * Represents a tranSMART user.
  */
 public interface User {
 
     /**
      * An numerical identifier for the user.
-     *
+     * @deprecated Use {@code getUserame()} instead to identify the user.
```
gijskant: Typo. You can also link to the function: `{@link #getUsername()}`.

transmart-core-db/grails-app/services/org/transmartproject/db/querytool/PatientSetQueryBuilderService.groovy
```diff
@@ -44,6 +45,8 @@ class PatientSetQueryBuilderService {
 
     def sessionFactory
 
+    AccessControlChecks accessControlChecks
```
gijskant: Can you annotate this with `@Autowired`?

transmart-core-db/grails-app/services/org/transmartproject/db/tree/TreeCacheService.groovy
```diff
@@ -131,30 +130,35 @@ class TreeCacheService {
...
         List<I2b2Secure> i2b2Nodes = criteria.getExecutableCriteria(sessionFactory.currentSession).list()
-        def t2 = new Date()
-        log.info "Found ${i2b2Nodes.size()} nodes. Query took ${t2.time - t1.time} ms."
-
-        List<TreeNode> forest = buildForest(i2b2Nodes)
-        def t3 = new Date()
-        log.debug "Forest growing took ${t3.time - t2.time} ms."
+        stopWatch.stop()
+        stopWatch.start('Filtering nodes user has access to.')
+        List<I2b2Secure> accessibleI2b2Nodes = i2b2Nodes.findAll { I2b2Secure i2b2Secure ->
```
gijskant: Logically fetching all and filtering on the onces you have access to is the same as fetching the once you have access to, but I am really in favour of have both functionalities:
- checking for an entity, if the user has sufficient permissions for it;
- fetching all entities (of a certain type/matching certain constraints) that the user has sufficient permissions for.

gijskant: The first one is currently available in the `AccessControlChecks`/`User` class, the second one seems appropriate for specific resources, such as this (for tree nodes).

transmart-core-db/grails-app/services/org/transmartproject/db/tree/TreeService.groovy
```diff
@@ -94,32 +90,19 @@ class TreeService implements TreeResource {
...
-            List<String> tokens = [Study.PUBLIC, 'EXP:PUBLIC'] + getStudyTokens(user)
-            criteria = criteria.add(Restrictions.in('secureObjectToken', tokens))
+        I2b2Secure i2b2Secure = (I2b2Secure) I2b2Secure.createCriteria().get {
+            eq('fullName', rootKey)
         }
-
-        I2b2Secure root = criteria.getExecutableCriteria(sessionFactory.currentSession).uniqueResult() as I2b2Secure
-        if (!root) {
+        assert i2b2Secure : "Node with path ${rootKey} is not found."
+        if (accessControlChecks.canPerform(user, ProtectedOperation.WellKnownOperations.SHOW_SUMMARY_STATISTICS, i2b2Secure)) {
```
gijskant: We actually want to throw the same error when the node does not exist as when the user does not have access.

gijskant: By the way, why change the name from `root` to `i2b2Secure`? Is it more descriptive of what we are fetching?

transmart-core-db/src/integration-test/groovy/org/transmartproject/db/accesscontrol/AccessControlChecksSpec.groovy
```diff
@@ -0,0 +1,60 @@
...
+import org.transmartproject.db.ontology.I2b2Secure
+import org.transmartproject.db.user.User
+import spock.lang.Specification
+
+import static org.transmartproject.core.users.ProtectedOperation.WellKnownOperations.READ
+import static org.transmartproject.core.users.ProtectedOperation.WellKnownOperations.SHOW_SUMMARY_STATISTICS
+
+@Rollback
+@Integration
+class AccessControlChecksSpec extends Specification {
```
gijskant: I like this test suite!

PR #240, opened 2018-06-04
Fix PostgreSQL limitation for IN query

ewelinagr:
Similar as for Oracle (limit = 1000), but the limit for PostgreSQL is 32767 items in 'IN' query

[TMT-155](https://jira.thehyve.nl/browse/TMT-155)

### DIFF
--- a/transmart-core-db/grails-app/services/org/transmartproject/db/support/InQuery.groovy
+++ b/transmart-core-db/grails-app/services/org/transmartproject/db/support/InQuery.groovy
@@ -11,11 +11,15 @@ import org.hibernate.criterion.Criterion
 import org.hibernate.criterion.Disjunction
 import org.hibernate.criterion.Restrictions
 import org.transmartproject.core.exceptions.InvalidRequestException
+import org.transmartproject.db.support.DatabasePortabilityService.DatabaseType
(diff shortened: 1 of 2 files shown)


transmart-core-db/grails-app/services/org/transmartproject/db/support/InQuery.groovy
```diff
@@ -25,12 +29,18 @@ import static org.transmartproject.db.support.DatabasePortabilityService.Databas
  */
 class InQuery {
 
-    public static final int MAX_LIST_SIZE = 1000
```
gijskant: 1000 will also work for PostgeSQL, so why make this distinction?

ewelinagr: I thought it is better for the performance to have less 'OR' clauses, so it would be better to have larger 'IN', if possible for PostgreSQL. However, I haven't tested if it's working better, so if you know there is no big difference, I'll change it.

gijskant:
I am closing the PR, because we need a different solution to the problem of [TMT-155](https://jira.thehyve.nl/browse/TMT-155).
The PostgreSQL driver does not accept more that 32,767 parameters for a query (not per `in` operator).
Looks like we need to split up the query into multiple queries or save the list of identifiers into a (temporary) table and refer to that table in the query.

PR #241, opened 2018-06-06
Add support of OIDC identity provider (keycloak) to transmart.

forus:


transmart-api-server/build.gradle
```diff
@@ -28,8 +28,9 @@ dependencies {
     compile project(':transmart-rest-api')
     compile project(':transmart-notifications')
 
-    compile "org.springframework.boot:spring-boot-starter-security"
```
forus: By removing this dependency we started to rely on transitive dependency in the code. Not desired scenario. I am going to fix this.

transmart-rest-api/src/main/groovy/org/transmartproject/rest/user/AuthContext.groovy
```diff
@@ -19,31 +18,11 @@ import org.transmartproject.core.users.UsersResource
...
-            log.debug("Principal is a '${userPrincipal}' string. Interpreting it as username.")
-            username = userPrincipal
-        } else if (userPrincipal.hasProperty('username')) {
-            log.debug("Principal has ${userPrincipal.class} type and has a username field.")
-            username = userPrincipal.username
-        } else {
-            throw new AccessDeniedException("Unsuported principal type: ${userPrincipal.class}.")
-        }
-        usersResourceService.getUserFromUsername(username)
+        usersResource.getUserFromPrincipal(SecurityContextHolder.context.authentication)
```
forus: this is single line that makes rest-api plugin depend on spring security library.
Considering to use `request.getuserprincipal()` instead and remove the dependency.

transmart-api-server/build.gradle
```diff
@@ -28,8 +28,9 @@ dependencies {
...
     compile "org.springframework.security.oauth:spring-security-oauth2"
```
forus: still wondering what the best strategy to pick the library version would be.
I would like to investigate whether spring boot bom (AFAIK used by grails implicitly?) has a version for this library. I don't know which version it would pick if bom does not contain anything.

transmart-api-server/src/main/groovy/org/transmartproject/api/server/user/KeycloakUserResourceService.groovy
```diff
@@ -0,0 +1,136 @@
...
+    @Override
+    List<User> getUsersWithEmailSpecified() {
+        throw new UnsupportedOperationException()
+    }
+
+    @Override
+    User getUserFromPrincipal(Principal principal) {
+        assert principal instanceof OAuth2Authentication
+        def authentication = principal.userAuthentication
+        assert authentication: 'User is not authenticated.'
```
gijskant: Why not throw an exception (in `transmart-core-api`), to be caught by, e.g., [BusinessExceptionResolver](https://github.com/thehyve/transmart-core/blob/dev/transmart-rest-api/src/main/groovy/org/transmartproject/rest/http/BusinessExceptionResolver.groovy)?

transmart-api-server/src/main/groovy/org/transmartproject/api/server/user/KeycloakUserResourceService.groovy
```diff
@@ -0,0 +1,136 @@
...
+
+        createUser(username, realName, email, admin, accessStudyTokenToOperations)
+    }
+
+    private User createUser(final String username,
+                            final String realName,
+                            final String email,
+                            final boolean admin,
+                            final Multimap<String, ProtectedOperation> accessStudyTokenToOperations) {
+        new User() {
```
gijskant: I don't really like this anonymous class style. Why not just create a regular class for this?

transmart-api-server/src/main/groovy/org/transmartproject/api/server/user/KeycloakUserResourceService.groovy
forus: core-api should not know anything about spring security. But we could still throw `AccessDeniedException` from here and it should be caught by `BusinessExceptionResolver`

transmart-api-server/src/main/groovy/org/transmartproject/api/server/user/KeycloakUserResourceService.groovy
```diff
@@ -0,0 +1,136 @@
...
+            }
+
+            @Override
+            String getEmail() {
+                email
+            }
+
+            @Override
+            boolean canPerform(ProtectedOperation operation, ProtectedResource protectedResource) {
+                authorisationChecks.canPerform(this, operation, protectedResource)
```
gijskant: Precisely why I think: why would you ask that to the user?

forus: That's legacy. The method has been deprecated in the core-api's `User` interface recently.

transmart-api-server/src/main/groovy/org/transmartproject/api/server/user/KeycloakUserResourceService.groovy
forus: We probably should do this. How `canPerform(...)` method implementation would look like then?

transmart-api-server/src/main/groovy/org/transmartproject/api/server/user/KeycloakUserResourceService.groovy
forus: What we could do though. Remove all usages of the method in the code.

transmart-api-server/grails-app/init/org/transmartproject/api/server/ResourceServerConfiguration.groovy
```diff
@@ -0,0 +1,28 @@
...
+import org.transmartproject.api.server.user.keycloak.SubPrincipalExtractor
+
+@Configuration
+@EnableResourceServer
+class ResourceServerConfiguration extends ResourceServerConfigurerAdapter {
+
+    @Override
+    void configure(HttpSecurity http) throws Exception {
+        http.authorizeRequests()
+                .antMatchers(HttpMethod.OPTIONS, "/**").permitAll()
```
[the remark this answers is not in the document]
forus: mmm. That we would need to repeat these filters in two places.

transmart-core-db/src/main/groovy/org/transmartproject/db/accesscontrol/AccessControlChecks.groovy
```diff
@@ -377,4 +369,17 @@ class AccessControlChecks {
         res
     }
 
+    /**
+     * Groovy supports multiple dispatch. So other {@see canPerform} method has to be called.
+     * This class does not have to be @CompileStatic in order for this mechanism to work.
```
gijskant: Would be great, though, to have as much `@CompileStatic` as possible in the code (although groovy has some weird behaviour with inheritance).

transmart-rest-api/src/main/groovy/org/transmartproject/rest/user/AuthContext.groovy
gijskant: And what if you don't have a request? E.g., asynchronous jobs or tests?

transmart-api-server/keycloak.md
```diff
@@ -0,0 +1,72 @@
...
+    -Roles
+    
+        ROLE_ADMIN
+        
+    -Study permissions
+    
+        SHOW_SUMMARY_STATISTICS|EXP:SCSCP
+        READ|EXP:SCSCP
+        
+    The convention is to represent study permissions as `operation|study token`
```
gijskant: Let's discuss if `permisson|study token` is not a better approach.

forus: Yes, `Permission` seems to be a good candidate for this.

forus:
The current implementation (via `UserInfoTokenServices`) using the access token makes a request to `userInfoUri` to get the user info. Despite the fact that the token already contains all necessary claims such as roles.
We could use just the information supplied with the token instead. For that, we can enhance `DefaultTokenServices` like Podium did instead.
https://github.com/thehyve/podium/blob/dev/podium-gateway/src/main/java/nl/thehyve/podium/config/MicroserviceSecurityConfiguration.java

application.yml
```diff
@@ -0,0 +1,114 @@
+---
```
ewelinagr: why is this config added on the main level, looks like a copy of `transmart-api-server/application.yaml`?

forus: you are right. it should not end up here.

transmart-core-db/src/main/groovy/org/transmartproject/db/accesscontrol/AccessControlChecks.groovy
```diff
@@ -53,14 +56,11 @@ import static org.transmartproject.db.ontology.AbstractAcrossTrialsOntologyTerm.
...
- * Right now these are implemented in this same class.
- * If this list gets bigger or needs to be pluggable by other plugins, then
- * this should be refactored into a different solution (e.g. each check
- * implemented in a different Spring bean).
+ * It implements checks for 17.1 data model and legacy one.
  */
 @Slf4j
 @Component
-class AccessControlChecks {
+class AccessControlChecks implements AuthorisationChecks, MultidimQueryAuthorisationChecks {
```
gijskant: Can we add `@CompileStatic` here?

transmart-core-api/src/main/groovy/org/transmartproject/core/users/AuthorisationChecks.groovy
```diff
@@ -0,0 +1,45 @@
+package org.transmartproject.core.users
+
+import org.transmartproject.core.ontology.OntologyTerm
+import org.transmartproject.core.ontology.Study
+import org.transmartproject.core.querytool.QueryDefinition
+import org.transmartproject.core.querytool.QueryResult
+
+interface AuthorisationChecks {
```
gijskant: Please name this `LegacyAuthorisationChecks` to signal that this is for the legacy model (based on study tree nodes).

transmart-core-api/src/main/groovy/org/transmartproject/core/users/MultidimQueryAuthorisationChecks.groovy
```diff
@@ -0,0 +1,25 @@
+package org.transmartproject.core.users
+
+import org.transmartproject.core.ontology.MDStudy
+
+/**
+ * 17.1 data model (study enitity, hypercube, constraints language) authorisation checks.
+ */
+interface MultidimQueryAuthorisationChecks {
```
gijskant: Please name this `AuthorisationChecks` to convey that this is the actual type people should use.

transmart-core-api/src/main/groovy/org/transmartproject/core/users/PatientDataAccessLevel.groovy
```diff
@@ -0,0 +1,35 @@
...
+ * Describes a level of read access to the patient data.
+ *
+ * Enum values declaration order specifies {@link Enum#compareTo(PatientDataAccessLevel acclvl)} how high is the access level.
+ * COUNTS_WITH_THRESHOLD < SUMMARY < MEASUREMENTS
+ */
+enum PatientDataAccessLevel {
+
+    /**
+     * Only access to patient counts, with resolution limited by the configured
+     * threshold. {@see Config}.
```
ewelinagr: which Config is that? 
maybe better to use `@link`

forus: we don't have this implemented yet. I'm going to remove the reference.

transmart-core-api/src/main/groovy/org/transmartproject/core/users/MultidimQueryAuthorisationChecks.groovy
forus: sounds good. implemented.

transmart-core-db/src/main/groovy/org/transmartproject/db/accesscontrol/AccessControlChecks.groovy
forus: Makes sense.

forus:
Points about the dependencies and the OAuth workflow will be taken into https://github.com/thehyve/transmart-core/pull/246

PR #245, opened 2018-06-13
Fix notifications plugin config

ewelinagr:
- rename 'application.yml' to 'plugin.yml' according to https://docs.grails.org/3.2.3/guide/plugins.html
- fix configuration formatting

Other plugins of transmart-api-server should be also checked whether they do not require similar modification. It is described in: https://jira.thehyve.nl/browse/TMT-276

### DIFF
--- a/transmart-notifications/grails-app/conf/plugin.yml
+++ b/transmart-notifications/grails-app/conf/plugin.yml
@@ -1,18 +1,4 @@
-grails:
-    mail:
-        host: "localhost:25"
-        'default':
-            from: "<default_email>"
-        port: 465
-        username: "<username>"
-        password: "<password>"
-        props:
-          "mail.smtp.auth":"true"
-          "mail.smtp.ssl.enable": "true"
-          "mail.smtp.socketFactory.port":"465"
-          "mail.smtp.socketFactory.class":"javax.net.ssl.SSLSocketFactory"
-          "mail.smtp.socketFactory.fallback":"false"
-
+---
 # max number of query sets returned in a subscription email
 org.transmart.notifications.maxNumberOfSets: 20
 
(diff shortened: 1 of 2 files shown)


transmart-notifications/grails-app/conf/plugin.yml
```diff
@@ -1,20 +1,7 @@
...
-          "mail.smtp.ssl.enable": "true"
-          "mail.smtp.socketFactory.port":"465"
-          "mail.smtp.socketFactory.class":"javax.net.ssl.SSLSocketFactory"
-          "mail.smtp.socketFactory.fallback":"false"
-
+---
 # max number of query sets returned in a subscription email
-org.transmart.notifications.maxNumberOfSets: 20
+notifications:
+    maxNumberOfSets: 20
```
gijskant: Why not put these in the `org.transmart` namespace?

transmart-notifications/grails-app/services/org/transmart/notifications/QuerySetSubscriptionMailService.groovy
```diff
@@ -22,7 +23,8 @@ import org.transmartproject.core.userquery.SetType
 @Slf4j
 class QuerySetSubscriptionMailService {
 
-    def grailsApplication
+    @Value('${org.transmart.notifications.maxNumberOfSets}')
+    def maxNumberOfSets
```
gijskant: Why not just add type `Integer` here then?

ewelinagr: I did not specify it as an Integer, because when this property is not set, not found or it is set in some incorrect way, it fails with an ugly stacktrace in logs. With this assertion the error message looks better. I can also add a different way of error handling if you think it would be better.

PR #246, opened 2018-06-13
Implement fetching users and roles from keycloak rest-api

ewelinagr:
Simple implementation to get users and roles from Keycloak using RestTemplate and an "offline_token". 
Includes a documentation of how to generate the token. 

Keycloak configuration requires some cleanup, mainly the KeycloakRestTemplate related settings.


transmart-api-server/grails-app/conf/application.yml
```diff
@@ -109,19 +109,10 @@ server:
...
-#    realm: transmart   # our keycloak realm this ressource service should be based in
-#    auth-server-url: http://localhost:8080/auth  # the url to our sso server (keycloak)
-#    ssl-required: external  # no ssl is ok for demo  - do not use this in production
-#    resource: tm_client  # the realm-client which accessing clients should be authenticate with
-#    use-resource-role-mappings: true # use this if you manage your roles in client scope not in realm scope
-#    public-client: false # we do not write a web-app - so no login page and redirects are necessary
-#    bearer-only: true  # authentification via accesstoken is fine
-#    credentials:
-#        secret: 7034b121-fdf1-4188-9ba2-c66bee4b9867
+            clientId: <your-client-id>
```
forus: hm, is it sufficient to know only the clientId to make transmart functioning as the resource servers?
It does not have a URL to make a back channel call to the keycloak. Which is fine. But it does not seem to have means to get publick key of the keycloak instance.

transmart-api-server/grails-app/services/org/transmart/api/server/KeycloakResourceService.groovy
```diff
@@ -0,0 +1,73 @@
+package org.transmart.api.server
+
+import grails.transaction.Transactional
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.http.*
+import org.springframework.util.LinkedMultiValueMap
+import org.springframework.util.MultiValueMap
+import org.springframework.web.client.RestTemplate
+
+@Transactional
```
forus: it does not seem to be needed.

transmart-api-server/keycloak.md
```diff
@@ -59,6 +59,23 @@ Set the fields as following:
...
+
+In order to access Keycloak offline (e.g. by offline quartz jobs from transmart-notifications), an offline_token has to be configured. To get the token on behalf of a user, the user needs to have the role mapping for the realm-level: `"offline_access"`.
+
+    curl \
+      -d 'client_id=<CLIENT_ID>' \
+      -d 'username=<USERNAME>' \
+      -d 'password=<PASSWORD>' \
+      -d 'grant_type=password' \
+      -d 'scope=offline_access' \
+      'https://YOUR_KEYCLOAK_SERVER_HOST/auth/realms/YOUR_REALM/protocol/openid-connect/token'
```
forus: 👍

transmart-api-server/src/main/webapp/WEB-INF/keycloak.json
```diff
@@ -0,0 +1,10 @@
+{
+  "realm": "transmart",
```
forus: we don't need this one anymore, right?

transmart-api-server/src/test/groovy/org/transmartproject/api/server/user/KeycloakUserResourceServiceSpec.groovy
```diff
@@ -69,8 +72,73 @@ class KeycloakUserResourceServiceSpec extends Specification {
...
+        result[0].realName == "${keycloakMockUsers[0].firstName} ${keycloakMockUsers[0].lastName}"
+        result[0].email == keycloakMockUsers[0].email
+        result[0].studyTokenToAccessLevel == ["STUDY1_TOKEN": OWN]
+
+        result[1].admin == false
+        result[1].username == keycloakMockUsers[1].username
+        result[1].realName == "${keycloakMockUsers[1].firstName} ${keycloakMockUsers[1].lastName}"
+        result[1].email == keycloakMockUsers[1].email
+        result[1].studyTokenToAccessLevel == [:]
+
```
forus: 👍

transmart-api-server/build.gradle
```diff
@@ -28,6 +28,7 @@ dependencies {
     compile project(':transmart-rest-api')
     compile project(':transmart-notifications')
 
+    compile group: 'cglib', name: 'cglib-nodep', version: '3.2.6'
```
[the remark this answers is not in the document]
ewelinagr: this one is to be able to create mocks of non-interface types. Apparently it is not possible by default in Spock.  
I used this mocking here: https://github.com/thehyve/transmart-core/pull/246/commits/f163a8bfd3af1cd26c9b0a9e4e91be5670a49d45#diff-087ab0abd06f51677a80623f5e5ee8beR116

transmart-api-server/grails-app/services/org/transmart/api/server/KeycloakResourceService.groovy
ewelinagr: ah yes, it was added automatically, when I created a new server, I'll remove it

transmart-api-server/grails-app/conf/application.yml
ewelinagr: I am not sure, everything seems working only with this and using also the offlineToken, when doing a rest call to keycloak, but maybe there is a case, where we need also the secret.

ewelinagr: but maybe it is still using the keycloak.json config, I did not remove the KeycloakRestTemplate settings.

transmart-api-server/src/main/webapp/WEB-INF/keycloak.json
ewelinagr: as I mentioned above, I did not remove all the KeycloakRestTemplate settings, so maybe it is still using it for the GB calls. I will test if it still works after removing this config.

transmart-api-server/build.gradle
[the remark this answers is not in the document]
forus: ah, I see. https://stackoverflow.com/questions/10664182/what-is-the-difference-between-jdk-dynamic-proxy-and-cglib

It has to have `test` scope then.

Alternatively, we could introduce an interface. That does not seem like a bad idea.

PR #258, opened 2018-06-25
Update e2e tests to work with keycloak

ewelinagr:


ewelinagr:
Still have some tests failing, so fixing is still in progress

ewelinagr:
All tests are green again!!!

transmart-data/test_data/keycloak/keycloak-test-resource.json
```diff
@@ -0,0 +1,1542 @@
+{
+  "id" : "test",
```
forus: what about having this realm settings as part of the keycloak libery test resources?

I will add this point to the following ticket's description https://jira.thehyve.nl/browse/TMT-261

ewelinagr:
@forus Configuration of e2e is updated now. Tests pass for both transmart-api-server and transmart-server.

PR #259, opened 2018-06-27
Data table tests

gijskant:
- Refactor data table code
- Reorganised and added tests

I don't know if the data table export (still) works.


transmart-core-api/build.gradle
```diff
@@ -5,9 +5,8 @@ dependencies {
     compile 'org.codehaus.groovy:groovy-all:2.4.7'
     compile 'javax.validation:validation-api:1.1.0.Final'
     testCompile 'junit:junit:4.12'
-    testRuntime 'org.hamcrest:hamcrest-core:1.3',
-            'org.hamcrest:hamcrest-library:1.3'
-    // https://mvnrepository.com/artifact/org.hamcrest/hamcrest-all
+    testRuntime 'org.hamcrest:hamcrest-core:1.3'
+    testRuntime 'org.hamcrest:hamcrest-library:1.3'
     compile group: 'org.hamcrest', name: 'hamcrest-all', version: '1.3'
```
ewelinagr: Do we need hamcrest-all if we have hamcrest-core and hamcrest-library? If yes, it could be changed to testCompile. I know it's not a part of this PR, just noticed that :)

PR #263, opened 2018-07-02
Fix error in cell order of data table TSV export.

gijskant:


transmart-rest-api-e2e/src/test/groovy/tests/rest/v2/DataExportSpec.groovy
```diff
@@ -43,7 +44,7 @@ class DataExportSpec extends RESTSpec {
         request.path = "$PATH_DATA_EXPORT/job"
         request.query = [name: name]
         response = post(request)
-        responseData = response.exportJob
+        responseData = ExportJob.from(response.exportJob)
```
ewelinagr: It looks like there are more places in this test class that could be replaced like this, e.g. [line 183](https://github.com/thehyve/transmart-core/pull/263/files#diff-d7d988d63ef259059b18505777b79275R183)

gijskant:
@ewelinagr Code now okay?

PR #265, opened 2018-07-03
Enable serving of OpenAPI documentation with transmart-api-server.

gijskant:
Fixes [TMT-314](https://jira.thehyve.nl/browse/TMT-314)

### DIFF
--- a/transmart-api-server/grails-app/controllers/org/transmartproject/api/server/ApiServerUrlMappings.groovy
+++ b/transmart-api-server/grails-app/controllers/org/transmartproject/api/server/ApiServerUrlMappings.groovy
@@ -1,0 +1,10 @@
+package org.transmartproject.api.server
+
+class ApiServerUrlMappings {
+
+    static mappings = {
+        "/"(redirect: "/open-api/index.html")
+        "/open-api"(redirect: "/open-api/index.html")
+    }
+
+}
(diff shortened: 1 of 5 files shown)


gijskant:
@forus Could you have a look at this one?

transmart-api-server/grails-app/services/org/transmartproject/api/server/StartupService.groovy
```diff
@@ -0,0 +1,20 @@
+package org.transmartproject.api.server
+
+import grails.util.Holders
+import groovy.transform.CompileStatic
+import org.springframework.boot.CommandLineRunner
+
+@CompileStatic
+class StartupService implements CommandLineRunner {
```
forus: Is it called when the application running from war file?

transmart-api-server/grails-app/controllers/org/transmartproject/api/server/ApiServerUrlMappings.groovy
```diff
@@ -0,0 +1,10 @@
+package org.transmartproject.api.server
+
+class ApiServerUrlMappings {
+
+    static mappings = {
+        "/"(redirect: "/open-api/index.html")
```
forus: would it always redirect independently what `Accept` value specified in the header?

PR #247, opened 2018-06-19
Extend API logging 

ewelinagr:
As a system administrator, I want to be able to see a log of relevant calls to the REST API (TMT-181):
TMT-180 - Add audit interceptors to log data requests for auditing
TMT-251 - Add saving API call logs to the file, fix an audit logger configuration

PR #266, opened 2018-07-04
Temporarily ignore randomly failing ChildProcessAppender test

ewelinagr:
@gijskant @forus There is one rest-api test for ChildProcessAppender (I kept it from a previous implementation that was using log4j specific libraries) that is failing randomly on dev (Travis). I am not sure why, because it is passing locally and was passing for the PR on Travis: https://github.com/thehyve/transmart-core/pull/247. Started failing after merge of this PR. I restarted rest-api tests couple of times and it is passing again. I need some time to investigate it, so as not to break future builds I would like to temporarily ignore it until I fix it.

### DIFF
--- a/transmart-rest-api/src/test/groovy/org/transmartproject/rest/logging/ChildProcessAppenderTests.groovy
+++ b/transmart-rest-api/src/test/groovy/org/transmartproject/rest/logging/ChildProcessAppenderTests.groovy
@@ -25,6 +25,7 @@ import org.junit.Rule
 import org.junit.rules.TemporaryFolder
 import ch.qos.logback.classic.Level
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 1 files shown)


transmart-rest-api/src/test/groovy/org/transmartproject/rest/logging/ChildProcessAppenderTests.groovy
```diff
@@ -101,6 +102,7 @@ class ChildProcessAppenderTests extends Specification {
         thrown ChildFailedException
     }
 
+    @Ignore
```
gijskant: Can you also document here in the code, why the test is being ignored temporarily?

ewelinagr:
@gijskant done

PR #267, opened 2018-07-05
Improve a way of disabling query subscription

ewelinagr:
Disabling of the subscription is here based on [ `enabled` property in TransmartNotificationsGrailsPlugin](https://github.com/thehyve/transmart-core/blob/d7b48a6df2f8dca9922fa2f94fb6537c8b71ef33/transmart-notifications/src/main/groovy/org/transmart/notifications/TransmartNotificationsGrailsPlugin.groovy#L33), which is inherited from grails.plugins.Plugin (that all the grails plugins extend). 

@gijskant do you think changes added here (together with already existing properties for disabling jobs and quartz plugin startup) are enough as for disabling query subscription functionality?

### DIFF
--- a/transmart-notifications/README.md
+++ b/transmart-notifications/README.md
@@ -38,7 +38,7 @@ For more information check [Grails Email Plugin documentation.](http://gpc.githu
 ### Other subscription settings:
 
 ```
-org.transmart.notifications:
+org.transmartproject.notifications:
     # enable daily and weekly notification jobs
     enabled: true
     # max number of query sets returned in a subscription email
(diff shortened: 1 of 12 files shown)


gijskant:
- Please rename namespace `org.transmart` to `org.transmartproject`
- Could you rebase this branch on `dev` and update the tests to the changed test setup
- ... we need some serious refactoring of the UserQueryController to properly validate constraints ...

@ewelinagr @forus

ewelinagr:
@gijskant Regarding the refactoring of UserQueryController, indeed the validation and parsing of query object need some refactoring. Are you ok with adding this in a separate PR, or better to do it within this one?

gijskant:
Yes, indeed, separate PR is fine, it was just a reminder for all of us.

PR #268, opened 2018-07-09
Fix the access check for concepts.

gijskant:


gijskant:
Fixes issue [TMT-325](https://jira.thehyve.nl/browse/TMT-325).

PR #271, opened 2018-07-11
Disable the "v1" API for transmart-api-server

ewelinagr:
Covers [TMT-298](https://jira.thehyve.nl/browse/TMT-298) and [TMT-328](https://jira.thehyve.nl/browse/TMT-328)


forus:
Nice job!

PR #272, opened 2018-07-12
Add rest api e2e tests for access token validation

ewelinagr:
Covers [TMT-327](https://jira.thehyve.nl/browse/TMT-327)

### DIFF
--- a/transmart-rest-api-e2e/src/test/groovy/base/RestHelper.groovy
+++ b/transmart-rest-api-e2e/src/test/groovy/base/RestHelper.groovy
@@ -15,8 +15,10 @@ class RestHelper {
             request.uri.path = requestMap.path
             request.uri.query = requestMap.query
 
-            if (!requestMap.skipOauth && AUTH_NEEDED) {
+            if (!requestMap.token && AUTH_NEEDED) {
                 testContext.getAuthAdapter().authenticate(getRequest(), (requestMap.user ?: DEFAULT_USER))
+            } else if (requestMap.token){
+                request.headers.'Authorization' = 'Bearer ' + requestMap.token
             }
 
             response.success { FromServer fromServer, body ->
@@ -45,8 +47,10 @@ class RestHelper {
             request.contentType = requestMap.contentType ?: ContentTypeFor.JSON
             request.body = requestMap.body
 
-            if (!requestMap.skipOauth && AUTH_NEEDED) {
+            if (!requestMap.token && AUTH_NEEDED) {
                 testContext.getAuthAdapter().authenticate(getRequest(), (requestMap.user ?: DEFAULT_USER))
+            } else if (requestMap.token){
+                request.headers.'Authorization' = 'Bearer ' + requestMap.token
             }
 
             response.success { FromServer fromServer, body ->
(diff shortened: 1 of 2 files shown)


transmart-rest-api-e2e/src/test/groovy/base/RestHelper.groovy
```diff
@@ -17,6 +17,8 @@ class RestHelper {
 
             if (!requestMap.skipOauth && AUTH_NEEDED) {
                 testContext.getAuthAdapter().authenticate(getRequest(), (requestMap.user ?: DEFAULT_USER))
+            } else if (requestMap.skipOauth && requestMap.token){
```
forus: Could we perhaps simplify the conditions by reducing the number of flags?
Are `skipOauth` and `AUTH_NEEDED` always opposite flags to each other? When they not?
Are `requestMap.skipOauth` and `requestMap.token` always have the same boolean value? When they not?

transmart-rest-api-e2e/src/test/groovy/tests/rest/access/AccessCheckSpec.groovy
```diff
@@ -0,0 +1,103 @@
...
+                token     : tokenForUnrestrictedUser
+        ])
+
+
+        when: "I fabricate the token by replacing original user data with a new user"
+        def tokenWithReplacedUser = replaceUserInToken(tokenForUnrestrictedUser, DEFAULT_USER)
+        def responseForNewUser = get([
+                path      : PATH_STUDIES,
+                acceptType: JSON,
+                user      : UNRESTRICTED_USER,
```
forus: Why do we have both `token` and `user`? `token` seems to be sufficient.
I don't fully understand where passing `UNRESTRICTED_USER` as the user and `DEFAULT_USER` in the token fails.

transmart-rest-api-e2e/src/test/groovy/tests/rest/access/AccessCheckSpec.groovy
```diff
@@ -0,0 +1,103 @@
...
+                token     : tokenWithReplacedUser,
+                statusCode: 401
+        ])
+
+        then: "The signature is verified as invalid, I am not authorized to access the resource"
+        responseForNewUser.error == "Unauthorized"
+        responseForNewUser.message == "Unable to authenticate using the Authorization header"
+    }
+
+    private String replaceUserInToken(String tokenForUnrestrictedUser, String newUser) {
```
forus: nice!

transmart-rest-api-e2e/src/test/groovy/base/RestHelper.groovy
ewelinagr: `requestMap.token` is a string, but I can replace the skipOauth one with it, because `skipOauth` is not used anymore in other cases that I added.

Regarding `AUTH_NEEDED` it was used for the previous flow with transmart-oauth2 - there was an option to switch off the authentication in the configuration file. Since we are still supporting authentication with transmart for tm-server, we should probably keep this `AUTH_NEEDED`. What do you think?

transmart-rest-api-e2e/src/test/groovy/tests/rest/access/AccessCheckSpec.groovy
ewelinagr: you are right, user is not needed in this case, I'll remove it from this map.  It fails on checking the signature on the keycloak side. It compares the signature from the last part of the token with the `payload` part of the token, where there is a user data.

transmart-rest-api-e2e/src/test/groovy/tests/rest/access/AccessCheckSpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+        tokenPayload.sub = getUsername(newUser)
+
+        String newTokenPayload = decodeMapToBase64String(tokenPayload)
+        tokenParts[1] = newTokenPayload
+
+        tokenParts.join(".")
+    }
+
+    private String decodeMapToBase64String(Map tokenPart) {
+        def tokenPartString = toJSON(tokenPart)
```
gijskant: In the other PR I removed the `toJSON` function. Could you use `new ObjectMapper().writeValueAsString(tokenPart)` instead? That would reduce the merge conflicts.

PR #273, opened 2018-07-13
Add access policy tests. Reorganising transmart-rest-api functional tests.

forus:


transmart-core-db/src/main/groovy/org/transmartproject/db/dataquery/highdim/parameterproducers/StandardAssayConstraintFactory.groovy
```diff
@@ -41,6 +42,9 @@ class StandardAssayConstraintFactory extends AbstractMethodBasedParameterFactory
     @Autowired
     QueriesResource queriesResource
 
+    @Autowired
+    UsersResource usersResource
```
gijskant: Looks like this one is not being used after all.

PR #279, opened 2018-07-17
 Save patient sets to database when creating cross table.

gijskant:
Fixes [TMT-243](https://jira.thehyve.nl/browse/TMT-243).
Includes #275.

### DIFF
--- a/transmart-rest-api/src/main/groovy/org/transmartproject/rest/marshallers/MDStudySerializationHelper.groovy
+++ b/transmart-rest-api/src/main/groovy/org/transmartproject/rest/marshallers/MDStudySerializationHelper.groovy
@@ -24,6 +24,7 @@ class MDStudySerializationHelper extends AbstractHalOrJsonSerializationHelper<St
                 id: study.id,
                 studyId: study.name,
                 bioExperimentId: study.bioExperimentId,
+                secureObjectToken: study.secureObjectToken,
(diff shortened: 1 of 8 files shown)


transmart-rest-api/src/main/groovy/org/transmartproject/rest/marshallers/MDStudySerializationHelper.groovy
```diff
@@ -24,6 +24,7 @@ class MDStudySerializationHelper extends AbstractHalOrJsonSerializationHelper<St
                 id: study.id,
                 studyId: study.name,
                 bioExperimentId: study.bioExperimentId,
+                secureObjectToken: study.secureObjectToken,
```
ewelinagr: Why is this needed to be exposed?

ewelinagr:
This will break some of the e2e tests, but I can fix them in a separate PR

transmart-rest-api/src/main/groovy/org/transmartproject/rest/marshallers/MDStudySerializationHelper.groovy
gijskant: Mainly for data loading scripts to determine which studies are non-public, for loading roles into Keycloak (and which token to use).

gijskant:
@ewelinagr For another ticket I am also fixing e2e tests, so I can fix this one as well.

PR #280, opened 2018-07-17
Rest api e2e fixes

gijskant:


gijskant:
The main purpose of this PR is the test named `aggregates for raw text concept does not contain values` in `AggregatesPerConceptSpec`. It is the result of investigating why, in Glowing Bear, we do not get categorical values for the `favouritebook` concept. The reason is: it is not a categorical concept, but a free text concept.

open-api/swagger.yaml
```diff
@@ -3004,11 +3013,11 @@ definitions:
             type: number
           max:
             type: number
-          average:
+          avg:
             type: number
           count:
             type: number
-          std_dev:
+          stdDev:
```
forus: is changing gb covered? is there ticket/PR?

transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/NumericalValueAggregates.groovy
```diff
@@ -1,11 +1,12 @@
 package org.transmartproject.core.multidimquery
 
+import groovy.transform.Canonical
 import groovy.transform.Immutable
 
 /**
  * Numerical values aggregates
  */
-@Immutable
+@Canonical
```
forus: Why?

transmart-rest-api-e2e/src/test/groovy/config/Config.groovy
```diff
@@ -5,24 +5,42 @@ package config
...
+        value
+    }
+
+    //$ gradle -DbaseUrl=http://transmart-pro-test.thehyve.net/ test
+    public static final String BASE_URL = getProperty('baseUrl', 'http://localhost:8081/', String)
 
      // Configure whether the currently used application for providing a REST API for a TranSMART supports the `v1` API.
      // In particular, if the application is transmart-api-server, this should be set to false.
-    public static final boolean IS_V1_API_SUPPORTED = false
+    public static final Boolean IS_V1_API_SUPPORTED = getProperty('v1Supported', Boolean.FALSE, Boolean)
```
forus: imho, in general, using primitive types is a safer choice. we escape NPEs.

transmart-rest-api-e2e/src/test/groovy/representations/AggregatesPerConcept.groovy
```diff
@@ -0,0 +1,41 @@
+package representations
+
+import groovy.transform.EqualsAndHashCode
+
+@EqualsAndHashCode
+class NumericalValueAggregates {
```
forus: more representation beans to the bucket ;) nice!

transmart-rest-api-e2e/src/test/groovy/tests/rest/v2/AggregatesPerConceptSpec.groovy
```diff
@@ -65,47 +54,72 @@ class AggregatesPerConceptSpec extends RESTSpec {
...
-        "POST" | _
-        "GET"  | _
+        def request = [
+                path      : PATH_AGGREGATES_PER_CONCEPT,
+                acceptType: JSON,
+                body: params
+        ]
+
+        when: 'I request aggregates for the favouritebook concept (of type Raw text)'
+        def response = post(request) as AggregatesPerConcept
```
forus: `as` would use groovy/grails converters. Should we stick with jackson?

transmart-rest-api-e2e/src/test/groovy/tests/rest/v2/CrossTableSpec.groovy
```diff
@@ -3,99 +3,103 @@ package tests.rest.v2
...
 
-        then: "Returned count for the first cell equals zero"
-        assert responseData3.rows.size() == 2
-        assert responseData3.rows[0] == [0, 5]
-        assert responseData3.rows[1] == [0, 3]
+        when: 'I request a cross table with a reference to a restricted concept in one of the constraints'
+        def errorResponse = post(request) as ErrorResponse
+
+        then: 'Access is denied'
+        errorResponse.httpStatus == HttpStatus.FORBIDDEN.value()
```
forus: Nice! I have another test in transmart-rest-api func. tests to remove `@Ignore` after the change.

open-api/swagger.yaml
gijskant: Interestingly, the REST API never did anything else, but both this document and Glowing Bear had it wrong. Glowing Bear does not use these aggregates, by the way.

transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/NumericalValueAggregates.groovy
gijskant: Because of consistency, see [CategoricalValueAggregates](https://github.com/thehyve/transmart-core/blob/dev/transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/CategoricalValueAggregates.groovy).

transmart-rest-api-e2e/src/test/groovy/config/Config.groovy
gijskant: In this case, the `getProperty` function will always return true or false. It is actually quite hard to get strict checking on parsing booleans...

gijskant:
All e2e tests are passing now.

PR #281, opened 2018-07-21
Refactor subscription and audit log

gijskant:
-  Refactor user query saving and subscription functionality. 
- Fix audit logging: include username, log failures.


transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryService.groovy
```diff
@@ -1,46 +1,192 @@
...
+        new ConstraintAccessChecker(currentUser, PatientDataAccessLevel.MEASUREMENTS,
+                authorisationChecks, legacyAuthorisationChecks, conceptsResource)
+                .build(constraint)
     }
 
     @Override
-    UserQuery get(Long id, User currentUser) {
-        Query query = Query.findByIdAndDeleted(id, false)
+    List<UserQueryRepresentation> list(User currentUser) {
+        def result = Query.createCriteria().list {
```
ewelinagr: So should we always use `createCriteria` over Gorm's `findBy`?

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryService.groovy
```diff
@@ -1,46 +1,192 @@
...
+    LegacyAuthorisationChecks legacyAuthorisationChecks
+
+    @Autowired
+    ConceptsResource conceptsResource
+
+    @Autowired
+    SessionFactory sessionFactory
+
+
+    private static void validateSubscriptionEnabled(Boolean subscribed, SubscriptionFrequency subscriptionFreq) {
```
ewelinagr: `Boolean subscribed` - ah ok, nice, it fixes the current issue for `subscribed == null`

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryService.groovy
gijskant: We do if we use `@CompileStatic`. This will change in Grails 3.3, which provides `@GrailsCompileStatic` which is aware of domain classes. Note that the argument of `list` is a closure, which is not compiled statically. (So for full static compilation, this is not sufficient).

gijskant:
Some of the guidelines I tried to follow:
- Keep controllers as simple as possible, because they are hard to test. Mainly provide (de)serialisation of the HTTP request and response;
- Avoid passing domain classes in the interface of a service (arguments or response). Having an interface in transmart-core-api is not sufficient, we bring in all the dynamic nature of GORM into the interface if the implementation is a domain class. This is not bad per se, but sometimes actions on these objects require a current session, and the validation you can do on these objects is limited to database constraints.

@ewelinagr @forus Do these sound reasonable, or are there important exceptions? (probably streaming lots of data, like in export, hypercube serialisation are exceptions...)

transmart-rest-api/grails-app/controllers/org/transmartproject/interceptors/ApiAuditInterceptor.groovy
```diff
@@ -30,24 +36,20 @@ class ApiAuditInterceptor {
...
     boolean after() {
         report("$controllerName request", eventMessage)
     }
 
     protected boolean report(String event, String eventMessage) {
-
-        accessLogService.report(
-                authContext?.user,
+        accessLogEntryResource.report(
+                authContext.user,
```
ewelinagr: there was an issue with null user in the log entry. Do you know what was the reason? It was not related to the authContext being null?

transmart-rest-api/src/func-test/groovy/org/transmartproject/rest/v2/ApiAuditInterceptorSpec.groovy
```diff
@@ -0,0 +1,358 @@
...
+import org.transmartproject.core.log.AccessLogEntryResource
+import org.transmartproject.interceptors.ApiAuditInterceptor
+import org.transmartproject.mock.MockAccessLogEntryResource
+import org.transmartproject.rest.BusinessExceptionController
+import spock.lang.Unroll
+
+import static org.springframework.http.HttpMethod.*
+
+@Slf4j
+class ApiAuditInterceptorSpec extends V2ResourceSpec {
```
gijskant: The main reason to switch from unit test to integration test is that the unit testing framework has difficulty to handle mandatory autowired dependencies. It would be nice if the grails testing framework offers easy mocking of injected beans, like the comination of JUnit, Spring testing and Mockito does.
It now checks if something is written to the log service, instead of matching of calls. It looks like the interceptor causes asynchronous behaviour, whithout the `sleep(1000)` I got errors. I will try if they are still needed.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
```diff
@@ -62,16 +76,17 @@ class UserQuerySetService implements UserQuerySetResource {
...
         }
 
-        for (query in userQueries) {
-
+        for (UserQuery query: userQueries) {
             List<QuerySetInstance> previousQuerySetInstances = getInstancesForLatestQuerySet(query.id)
-            ArrayList<Long> newPatientIds = getPatientsForQuery(query, currentUser)
+            User user = usersResource.getUserFromUsername(query.username)
+            def queryRepresentation = UserQueryService.toRepresentation(query)
+            List<Long> newPatientIds = getPatientsForQuery(queryRepresentation, user)
```
ewelinagr: this fixes [TMT-274](https://jira.thehyve.nl/browse/TMT-274)

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryService.groovy
ewelinagr: ok, thanks for the explanation, I will remember this next time

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
```diff
@@ -166,19 +182,19 @@ class UserQuerySetService implements UserQuerySetResource {
...
                 .add(Restrictions.eq('query.id', id))
-                .setProjection(Projections.max("createDate"))
-
-        // get the content for the most recent createDate and queryId
-        def session = sessionFactory.currentSession
-        QuerySet recent = session.createCriteria(QuerySet)
-                .add(Restrictions.eq("query.id", id))
-                .add(Property.forName( "createDate" ).eq(recentDate))
-                .uniqueResult()
+                .addOrder(Order.desc('createDate'))
```
ewelinagr: Why is ordering better then projection for the max 'createDate'?

transmart-rest-api/src/func-test/groovy/org/transmartproject/rest/v2/ApiAuditInterceptorSpec.groovy
gijskant: Removing the `sleep` still makes them all fail. Weird, because you are supposed to be able to alter the response in the `after()` function. I would then expect that the action has completed before sending the response.

transmart-rest-api/grails-app/controllers/org/transmartproject/interceptors/ApiAuditInterceptor.groovy
gijskant: Yes, adding `@Autowired` to the authContext seemed to solve that issue. Apparently the bean was not automatically injected.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
gijskant: Because you can do it in one query instead of two. First fetching the one with max date does a query on the query table, fetching the one with that date also. This solution does fetching the query and selecting the one with the max date in one query. It still only fetches the first, because of the `maxResults`.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryService.groovy
```diff
@@ -1,46 +1,192 @@
...
+    AuthorisationChecks authorisationChecks
+
+    @Autowired
+    LegacyAuthorisationChecks legacyAuthorisationChecks
+
+    @Autowired
+    ConceptsResource conceptsResource
+
+    @Autowired
+    SessionFactory sessionFactory
```
ewelinagr: Is sessionFactory needed here?

gijskant: Apparently not :) I will delete it.

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQueryService.groovy
```diff
@@ -1,46 +1,188 @@
...
+                checkConstraintAccess(ConstraintFactory.read(query.patientsQuery), currentUser)
+                if (!query.subscribed) {
+                    // This is a new subscription, an initial patient set needs to be generated
+                    newSubscription = true
+                }
+            }
+            query.subscribed = representation.subscribed
+        }
+        if (representation.subscriptionFreq != null) {
+            query.subscriptionFreq = representation.subscriptionFreq
```
ewelinagr: This field could be cleared when unsubscribing, but it is not that important, since it is not used anywhere, when not subscribed.

PR #285, opened 2018-07-25
Clean plugins and api-server conf.

forus:
api-server, rest-api, core-db-tests and core-db

plugin.yml gets merged with main application settings. application.yml of plugins doesn't.

it seems like non-grails settings placed in plugin.yml are not visible neither by the application nor by plugin itself.


transmart-core-db-tests/grails-app/conf/application.yml
```diff
@@ -1,135 +1,25 @@
...
-    groovy:
-        template:
-            check-template-location: false
-# Spring Actuator Endpoints are Disabled by Default
-endpoints:
-    enabled: false
-    jmx:
-        enabled: true
+      dataSource:
+        url: jdbc:h2:mem:testDb;MVCC=TRUE;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE;INIT=RUNSCRIPT FROM '../transmart-core-db/h2_init.sql'
```
ewelinagr: There is a small formatting mixup in this file with a number of spaces - sometimes 2, sometimes 4

transmart-rest-api/grails-app/conf/application.yml
```diff
@@ -0,0 +1,25 @@
...
+environments:
+    test:
+      hibernate:
+          cache:
+              queries: false
+              use_second_level_cache: true
+              use_query_cache: false
+              region.factory_class: org.hibernate.cache.ehcache.SingletonEhCacheRegionFactory
+
+      dataSource:
```
ewelinagr: same here, number of spaces is not equal everywhere

transmart-core-db-tests/grails-app/conf/application.yml
forus: thanks, I'll fix that.

forus:
I used 4 spaces instead of 2.

PR #286, opened 2018-07-25
Add a call to trigger sending of emails for query subscription

ewelinagr:
Covers [TMT-302](https://jira.thehyve.nl/browse/TMT-302)


build.gradle
```diff
@@ -98,7 +98,7 @@ subprojects { project ->
     apply plugin: 'idea'
 
     if (project.name =~ /^(biomart-domain|folder-management-plugin|mock-ontology-server|Rmodules|search-domain|transmart-core-db-tests|transmart-core-db|transmart-gwas-plugin|transmart-legacy-db|transmart-rest-api|transmartApp|transmart-server|transmart-oauth|transmart-solr-indexing|transmart-notifications|transmart-api-server)$/) {
-        if (project.name =~ /^(transmartApp|mock-ontology-server|transmart-server|transmart-solr-indexing|transmart-api-server)$/) {
+        if (project.name =~ /^(transmartApp|mock-ontology-server|transmart-server|transmart-solr-indexing|transmart-api-server|transmart-notifications)$/) {
```
gijskant: Why is this needed, it's doesn't have to run standalone, right?

transmart-notifications/grails-app/controllers/org/transmartproject/notifications/NotificationsUrlMappings.groovy
```diff
@@ -0,0 +1,12 @@
+package org.transmartproject.notifications
+
+class NotificationsUrlMappings {
+
+    static mappings = {
+        group "/v2", {
+            group "/admin", {
+                "/subscription/notify"(method: 'GET',controller: 'subscriptionMail', action: 'subscriptionNotify')
```
gijskant: You could also surround this with `if (Holders.config.org.transmartproject.notifications.endpointEnabled) {` (and add an entry in the config file) to be able to disable/enable this endpoint entirely, decause it is supposed to be for testing only.

transmart-notifications/grails-app/controllers/org/transmartproject/notifications/SubscriptionMailController.groovy
```diff
@@ -0,0 +1,36 @@
...
+     *
+     * @param frequency specifies whether the mail should be send to users
+     *        who subscribed for DAILY or WEEKLY subscription.
+     * @return {@link HttpStatus#OK};
+     *      or {@link HttpStatus#FORBIDDEN} when the user does not have an ADMIN role,
+     *      see request matcher configuration of the current application.
+     * @throws {@link InvalidArgumentsException} if frequency parameter is not valid.
+     */
+    def subscriptionNotify(@RequestParam('frequency')String frequency) {
+        SubscriptionFrequency subscriptionFrequency = frequency?.trim() ? SubscriptionFrequency.forName(frequency) : null
```
gijskant: Software development is a bit of a language game... looks to me that subscribing/subscription are one thing and sending notifications about updates another. I would prefer SubscriptionFrequency to be renamed to NotificationFrequency, etc to better reflect that distinction.

transmart-notifications/grails-app/controllers/org/transmartproject/notifications/SubscriptionMailController.groovy
```diff
@@ -0,0 +1,36 @@
+package org.transmartproject.notifications
+
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.http.HttpStatus
+import org.springframework.web.bind.annotation.RequestParam
+import org.transmartproject.core.exceptions.InvalidArgumentsException
+import org.transmartproject.core.userquery.SubscriptionFrequency
+
+class SubscriptionMailController {
```
gijskant: Code looks really nice. Simple and well documented. But perhaps this controller can be named NotificationMailController or NotificationController, because it does not mail subscriptions.

transmart-rest-api-e2e/src/test/groovy/tests/rest/v2/SubscriptionSpec.groovy
```diff
@@ -0,0 +1,74 @@
+package tests.rest.v2
+
+import base.RESTSpec
+
+import static base.ContentTypeFor.JSON
+import static config.Config.*
+
+/**
+ * These tests tequires transmart-notifications plugin to be enabled and properly configured.
```
gijskant: tequires -> requires

transmart-rest-api-e2e/src/test/groovy/tests/rest/v2/SubscriptionSpec.groovy
```diff
@@ -0,0 +1,74 @@
...
+    void 'test triggering email sending by a regular user is denied'() {
+        when: "I try to trigger email sending as default user"
+        def frequency = ""
+        def responseData = get([
+                path      : PATH_SUBSCRIPTION,
+                acceptType: JSON,
+                query     : [frequency: frequency],
+                user      : DEFAULT_USER,
+                statusCode: 403
+        ])
```
gijskant: You can now convert this to `ErrorResponse` so that we also have type checking in our test code.

transmart-api-server/src/main/groovy/org/transmartproject/api/server/server/SecurityConfig.groovy
```diff
@@ -62,6 +62,7 @@ class SecurityConfig extends KeycloakWebSecurityConfigurerAdapter {
         super.configure(http)
         http.authorizeRequests()
                 .antMatchers("/v1/*").denyAll()
+                .antMatchers("/v2/admin/**").hasRole('ADMIN')
```
gijskant: Nice!

transmart-notifications/build.gradle
```diff
@@ -1,17 +1,27 @@
 dependencies {
+    //Grails 3 dependencies
     compile "org.springframework.boot:spring-boot-starter-logging"
     compile "org.springframework.boot:spring-boot-autoconfigure"
     compile "org.grails:grails-core"
-    compile "org.grails:grails-dependencies"
+    compile "org.springframework.boot:spring-boot-starter-actuator"
```
gijskant: Are these all needed?

gijskant:
All changes are fine, but the dependencies in transmart-notifications/build.gradle do not seem needed.

transmart-notifications/grails-app/controllers/org/transmartproject/notifications/SubscriptionMailController.groovy
ewelinagr: true, I wanted it to be in line with QuerySetSubscriptionMailService, I guess more renaming is needed then

transmart-notifications/build.gradle
ewelinagr: these are the default (auto-generated) dependencies for grails rest-api profiled plugin (I removed only the hibernate ones), but you're right, probably some of them are not needed, I will do some more cleanup

transmart-notifications/grails-app/controllers/org/transmartproject/notifications/SubscriptionMailController.groovy
ewelinagr: The full renaming of SubscriptionFrequency would require changes in GB, should we do this before the release? I added some renaming to this controller and the added endpoint.

transmart-notifications/grails-app/controllers/org/transmartproject/notifications/NotificationsUrlMappings.groovy
ewelinagr: I tried this way, but adding this condition here was causing some issues:
```
Field urlMappings in org.grails.web.mapping.servlet.UrlMappingsErrorPageCustomizer required a bean of type 'grails.web.mapping.UrlMappings' that could not be found.
```
I disabled it in the controller.

PR #287, opened 2018-07-25
Change logback configuration of transmart-api-server with production mode

ewelinagr:
[TMT-293](https://jira.thehyve.nl/browse/TMT-293)

Configuration of logback is similar to the one for transmart-server with the difference that there is a different rolling policy. Instead of size only based, transmart-api will have a time-and-size policy - daily rollover, limited by size size of 100MB. 
There are also 2 optional parameters (currently commented out):
- maxHistory that controls the maximum number of archive files to keep, asynchronously deleting older files.
- totalSizeCap that controls the total size of all archive files.
We can set them if needed.

### DIFF
--- a/transmart-api-server/grails-app/conf/logback.groovy
+++ b/transmart-api-server/grails-app/conf/logback.groovy
@@ -3,8 +3,8 @@ import grails.util.BuildSettings
 import grails.util.Environment
 import org.springframework.boot.logging.logback.ColorConverter
 import org.springframework.boot.logging.logback.WhitespaceThrowableProxyConverter
-import org.transmartproject.rest.logging.ChildProcessAppender
 import org.transmartproject.rest.logging.ApiAuditLogJsonLayout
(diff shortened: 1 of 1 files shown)


transmart-api-server/grails-app/conf/logback.groovy
```diff
@@ -25,29 +24,54 @@ appender('STDOUT', ConsoleAppender) {
...
+                    '%logger{10} [%file:%line] %msg%n'
+        }
+        rollingPolicy(SizeAndTimeBasedRollingPolicy) {
+            // daily rollover
+            fileNamePattern = "${logDirectory}/transmart.%d{yyyy-MM-dd}.%i.log"
+            // size limit on the log files
+            maxFileSize = '100MB'
+            /* // optional parameters
+            // maxHistory controls the maximum number of archive files to keep, asynchronously deleting older files.
+            maxHistory = 30
```
gijskant: Maybe good to cap this to 9, such as in transmart-server.

PR #293, opened 2018-07-25
Use permissions as cache key instead of username.

gijskant:
Fixes [TMT-344](https://jira.thehyve.nl/browse/TMT-344).

- [x] Needs to be tested.


forus:
Currently, we iterate over users. That's the expensive part. What'd suit this change is iterating over unique combinations of `{#user.admin, #user.studyToPatientDataAccessLevel}`.

PR #294, opened 2018-07-26
Prevent writing empty rows to dimensions-data files during TSV export

ewelinagr:
[TMT-352](https://jira.thehyve.nl/browse/TMT-352)


ewelinagr:
The reason was that during the creation of the datatable currently it is possible to have [dim, [keys, elements]] elementMap for e.g. visit dimension, where key and elements are null (so empty table columns/rows). 
During the tsv export there are separate files per dimension created, which does not need to have empty rows.

ewelinagr:
If it should not be possible to have [dim, [keys, elements]] with null values [in the elementsMap of FullDataTable](https://github.com/thehyve/transmart-core/blob/dev/transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/FullDataTable.groovy#L60), I'll fix it in the FullDataTable class directly, but it seems to be issue only in the tsv export.

PR #297, opened 2018-07-30
Fix generating emails with empty updates for query subscription

ewelinagr:
TMT-362

### DIFF
--- a/transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
+++ b/transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
@@ -107,7 +107,7 @@ class UserQuerySetService implements UserQuerySetResource {
     List<UserQuerySetChangesRepresentation> getQueryChangeHistoryByUsernameAndFrequency(SubscriptionFrequency frequency,
                                                                                         String username,
                                                                                         Integer maxNumberOfSets) {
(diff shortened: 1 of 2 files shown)


transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
```diff
@@ -147,6 +147,7 @@ class UserQuerySetService implements UserQuerySetResource {
                 .add(Restrictions.eq('query.deleted', false))
                 .add(Restrictions.eq('query.subscribed', true))
                 .add(Restrictions.eq('query.subscriptionFreq', frequency))
+                .add(Restrictions.sizeGt('querySet.querySetDiffs', 0))
```
forus: You might prefer to use `isNotEmpty` instead `sizeGt`

According with the hibernate source code:
1. `isNotEmpty` generates sql like: `not exists (select 1 from <table> where ...)`
2. `sizeGt` : `0 < (select count(*) from <table> where ...)`

AFAIK SQL engines do optimizations of `exist`: they stop scanning the table to the end if there is at least one result.
https://blog.jooq.org/2016/09/14/avoid-using-count-in-sql-when-you-could-use-exists/

ewelinagr: ok, good one, I didn't think about `isNotEmpty`. I updated the PR

PR #299, opened 2018-07-30
Fix the query builder for when a user has no access to any study.

gijskant:
Fixes [TMT-363](https://jira.thehyve.nl/browse/TMT-363).


transmart-core-db-tests/src/integration-test/groovy/org/transmartproject/db/multidimquery/HibernateCriteriaQueryBuilderSpec.groovy
```diff
@@ -443,4 +448,54 @@ class HibernateCriteriaQueryBuilderSpec extends Specification {
...
+        def results = getList(criteria) as List
+
+        then: 'The result is empty'
+        results.empty
+    }
+
+    void 'test queries for user with no access to any study'() {
+        given: 'No access to any study'
+        setupHypercubeData()
+        QueryBuilder builder = HibernateCriteriaQueryBuilder.forStudies([])
```
forus: It seems to me like the previous test does not add anything new.
`HibernateCriteriaQueryBuilder` works with studies supplied via `forStudies(List)` method.
Both, no studies in the database or no studies user have access to conduct as an empty list of studies.

PR #302, opened 2018-08-02
Fix constraint validation for collections

ewelinagr:
- for value and field constraints, the validation should not reject collection values for collection operators.
- the "CONTAINS" operator should not be used as "IN" operator for non-string types.

Fixes: 
[TMT-131](https://jira.thehyve.nl/browse/TMT-131)
[TMT-359](https://jira.thehyve.nl/browse/TMT-359)


ewelinagr:
@gijskant I still see one inconsistency between the documentation and implementation. Should the operator between(<-->) be supported for "NUMERIC" value types? The documentation shows that it is rather an operator for a type equals "DATE", but int the code it is supported also for type "NUMERIC".

gijskant:
Support for numeric values is also fine, I guess.

PR #304, opened 2018-08-02
Include patients for summary-level access in set of all patients.

gijskant:
Fixes [TMT-369](https://jira.thehyve.nl/browse/TMT-369).


gijskant:
You can merge when okay.

PR #314, opened 2018-08-13
Change 'export data table with a modifier dimension' e2e test assertion

ewelinagr:
Test data loaded to Oracle db via groovy loading script does not always have 5 digits, see https://jira.thehyve.nl/browse/TMT-402


transmart-rest-api-e2e/src/test/groovy/tests/rest/v2/DataTableSpec.groovy
```diff
@@ -275,29 +275,27 @@ class DataTableSpec extends RESTSpec {
...
-                '52.00000',
-                'sample9',
-                '',
-                'sample7;sample8',
-                'sample6',
-                '380.00000;240.00000',
-                '28.00000'
-        ]
+        exportData[3][0] == '-63/TNS:63'
+        new BigDecimal(exportData[3][1]).compareTo(new BigDecimal('40.00000')) == 0
```
forus: maybe use `.compareTo(new BigDecimal('40'))` instead?

PR #316, opened 2018-08-14
Apply cache to the counts_per_concept call

ewelinagr:
[TMT-400](https://jira.thehyve.nl/browse/TMT-400)

### DIFF
--- a/transmart-core-db/src/integration-test/groovy/org/transmartproject/db/multidimquery/AggregateDataServicePgSpec.groovy
+++ b/transmart-core-db/src/integration-test/groovy/org/transmartproject/db/multidimquery/AggregateDataServicePgSpec.groovy
@@ -2,7 +2,6 @@
 
 package org.transmartproject.db.multidimquery
 
-import com.fasterxml.jackson.databind.ObjectMapper
 import grails.test.mixin.integration.Integration
 import grails.transaction.Rollback
 import org.springframework.beans.factory.annotation.Autowired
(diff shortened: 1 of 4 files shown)


forus:
Could you please add cleaning the cache to the `SystemService.clearCaches()` method?

ewelinagr:
@forus Done! Both for counts per concept and counts per studies (this one was also missing there) caches.

transmart-core-db/src/integration-test/groovy/org/transmartproject/db/multidimquery/AggregateDataServicePgSpec.groovy
```diff
@@ -198,4 +203,123 @@ class AggregateDataServicePgSpec extends Specification {
...
+        def admin = usersResource.getUserFromUsername('admin')
+        Constraint studyConstraint = new StudyNameConstraint(studyId: "EHR")
+
+        when: "I call countsPerConcept for study EHR"
+        aggregateDataService.countsPerConcept(studyConstraint, user)
+        def countsPerConceptCache = cacheManager.getCache(countsPerConceptCacheName)
+
+        then: "countsPerConcept cache contains counts for study EHR and non-admin user"
+        countsPerConceptCache.getNativeCache().size() == 1
+        countsPerConceptCache.getAllKeys()[0] == [new ObjectMapper().writeValueAsString(["type": "study_name", "studyId": "EHR"]),
```
forus: I'd say we don't need to check exact match of the keys here. It's other tests concern. Changing how the key is calculated or serialized would break many places instead of one.
IMHO checking that cache got a new entry after the call is sufficient. Testing it for other inputs is not necessary either.

hm, seems like `getNativeCache()` returns an object, by taking `size()` on it, we rely on the particular setup. But I don't see a better way at the moment (I am not sure how to inject a mock cache manager easily).

ewelinagr: true, tests are updated now

PR #317, opened 2018-08-14
Make BEFORE and AFTER constraint operators inclusive

ewelinagr:
[TMT-411](https://jira.thehyve.nl/browse/TMT-411)


transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/query/HibernateCriteriaQueryBuilder.groovy
```diff
@@ -372,14 +372,14 @@ class HibernateCriteriaQueryBuilder extends ConstraintBuilder<Criterion> impleme
...
-            case Operator.AFTER:
                 return Restrictions.gt(propertyName, value)
             case Operator.GREATER_THAN_OR_EQUALS:
+            case Operator.AFTER:
                 return Restrictions.ge(propertyName, value)
             case Operator.LESS_THAN:
-            case Operator.BEFORE:
                 return Restrictions.lt(propertyName, value)
             case Operator.LESS_THAN_OR_EQUALS:
+            case Operator.BEFORE:
```
forus: AFTER and BEFORE are expected to be exclusive. Unlike BETWEEN.

BETWEEN x AND y
expr2 less than or equal to expr1 AND expr1 less than or equal to expr3
https://docs.oracle.com/cd/B28359_01/server.111/b28286/conditions011.htm#SQLRF52147

a BETWEEN x AND y
is equivalent to
a >= x AND a <= y
https://www.postgresql.org/docs/10/static/functions-comparison.html

forus: let's fix how between is resolved instead

ewelinagr: but the BETWEEN operator is working as expected - is inclusive, the request in TMT-411 was to change the BEFORE and  AFTER operators to be inclusive too

forus: @wardweistra @jochemb I don't think it is a bug. Don't you think it's a mistake to make BEFORE/AFTER inclusive?

forus: Maybe we should make `LESS_THAN_OR_EQUALS` and `GREATER_THAN_OR_EQUALS` to work with dates.

ewelinagr: Ok, I am closing this PR and will create a new one with LESS_THAN_OR_EQUALS and GREATER_THAN_OR_EQUALS operators support for dates. There will be also some changes in GB.

PR #321, opened 2018-08-20
Fix the counts per study and concept for an empty database.

gijskant:


transmart-core-db/grails-app/services/org/transmartproject/db/clinical/MultidimensionalDataResourceService.groovy
```diff
@@ -468,11 +468,14 @@ class MultidimensionalDataResourceService extends AbstractDataResourceService im
         assaysByType.keySet()?.collect {it.dataTypeName}
     }
 
-    private List<MDStudy> selectStudiesWithDimensionSupport(Iterable<MDStudy> studies, DimensionImpl dimension) {
+    private List<MDStudy> selectStudiesWithDimensionSupport(Collection<MDStudy> studies, DimensionImpl dimension) {
+        if (studies.empty) {
```
forus: `if (studies) {` would escape possible NPE, nevertheless pass `null` is a bad idea.

PR #325, opened 2018-08-23
Improve email template

forus:


transmart-notifications/grails-app/conf/plugin.yml
```diff
@@ -12,8 +12,8 @@ org.transmartproject.notifications:
     # hh: Hour, range: 0-23;
     # mm: Minute, range: 0-59;
     dailyJobTriggerTime: 0-0
-    # Subject for the query subscription email
-    emailSubject: Test email for query subscription
+    # Client application name referred from email content
+    clientApplicationName: Glowing Bear
```
ewelinagr: could you add it to the transmart-notifications/README as well?

gijskant:
@forus Did you consider to actually use a template engine (e.g., Thymeleaf) for this instead of writing html in strings?

forus:
@gijskant it makes a lot of sense. Here is a ticket to improve the way it's implemented even further. https://jira.thehyve.nl/browse/TMT-454

PR #278, opened 2018-07-17
Counts with threshold tests and fixes.

forus:

PR #329, opened 2018-08-29
Proposal for restructuring the counts with threshold functionality.

gijskant:
Some aspects:
- organise static helper functions in a separate class
- split implementations for cross table and counts
- let one service implement the resource interface, because
  the services do different things and, hence, should not be
  implementing the same interface. (Probably the service without
  threshold should use different names for the implementing

Note:
To keep the unit tests for the threshold layer working, there
still needs to be separate beans, but ideally I think there
should only be one for counts and one for the cross table, instead
of two for both.


gijskant:
@forus @ewelinagr I have some doubts about the structure of the counts with threshold implementation. This PR serves as discussion piece, to think further about how to organise the services and interfaces.

gijskant:
Suggestions have been processed in #278.

PR #334, opened 2018-09-03
Fix user query related domain object mapping for async scan

ewelinagr:
The default map constructors were not working for asynchronous call.

Fixes [TMT-456](https://jira.thehyve.nl/browse/TMT-456)


transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
```diff
@@ -209,36 +202,35 @@ class UserQuerySetService implements UserQuerySetResource {
...
-                        changeFlag: ChangeFlag.REMOVED
-                )
+                QuerySetDiff setDiff = new QuerySetDiff()
+                setDiff.querySet = querySet
+                setDiff.objectId = it
+                setDiff.changeFlag = ChangeFlag.REMOVED
+                setDiff
             })
+
             querySet.save(flush: true)
```
gijskant: Do we also want to have `failOnError: true` here, to better capture errors?

ewelinagr: hmm, but then maybe better to catch the errors and just log some warnings not to fail the scan for all queries, when there is some data issue for one query?

PR #335, opened 2018-09-04
Add index on observation_fact including primary key and trial_visit_num.

gijskant:


forus:
@gijskant maybe we should make `ts-copy` not aware what exact indexes are there.  It should be just able to recreate indexes that were dropped.

What indexes to create could differ between transmart installations.

gijskant:
@forus That would be nice indeed, but the distinction between parent and child tables, for partitioning, makes that quite tricky.

gijskant:
@forus It is possible, but it would eliminate the possibility to execute dropping and restoring of indexes as a separate step, e.g.:
- Drop indexes
- Load study A
- Load study B
- ...
- Restore indexes

PR #336, opened 2018-09-04
Logging configuration update.

gijskant:
Previously, an error message would be written to stdout:
```
LOGBACK: No context given for ch.qos.logback.contrib.json.classic.JsonLayout@...
```


ewelinagr:
Was this the cause of the huge stacktrace produced in some cases?

gijskant:
No, but something I noticed during the investigation.

PR #338, opened 2018-09-06
Move rebuild_cache and clear_cache calls to SystemController

ewelinagr:
[TMT-460](https://jira.thehyve.nl/browse/TMT-460)


transmart-core-db/grails-app/services/org/transmartproject/db/support/SystemService.groovy
```diff
@@ -124,7 +134,7 @@ class SystemService implements SystemResource {
             'user query set scan': { ->
```
ewelinagr: `patientSetService.clearPatientSetIdsCache` above is done twice here - before and after refreshing the bitset materialized view (it is also a part of clearCaches()). Is it on purpose or I can remove one of them? @gijskant @forus

gijskant: One of them can go indeed.

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/SystemController.groovy
```diff
@@ -55,4 +56,66 @@ class SystemController {
...
+     * or 'stopped'.
+     * Example: <code>{"status": "running"}</code>.
+     *
+     * Only available for administrators.
+     *
+     * @return an object with a field <code>status</code> with value <code>running</code>
+     * or <code>stopped</code>.
+     *
+     */
+    def rebuildStatus() {
```
gijskant: Because you're using the same lock as for the update after data loading, it actually makes sense to have only one status call. You could populate the `updateStatus` object similarly as in that call, but then with a single task.

gijskant: Proposal here: #339.

ewelinagr: great, thank you! I will have a look

PR #340, opened 2018-09-06
Split aggregates_per_concept to two: aggregates_per_numerical_concept and aggregates_per_categorical_concept

forus:


ewelinagr:
There is an e2e test class `AggregatesPerConceptSpec`, is it still useful to have tests for deprecated call? Maybe better to remove it or change to use the new calls?

open-api/swagger.yaml
```diff
@@ -3125,32 +3174,56 @@ definitions:
     type: object
     properties:
       numericalValueAggregatesPerConcept:
+        $ref: '#/definitions/NumericalAggregates'
+      categoricalValueAggregatesPerConcept:
+        $ref: '#/definitions/NumericalAggregates'
```
ewelinagr: seems like it should be ` $ref: '#/definitions/CategoricalAggregates'` instead

forus: nice catch. thanks!

forus:
@ewelinagr we will remove e2e tests when we delete the call.
I thought about e2e tests for the new calls. But the functional tests in the `transmart-rest-api` seem to cover much logic already. I am not sure what e2e would add up.

PR #341, opened 2018-09-06
Upgrade Swagger UI to version 3.18.2

ewelinagr:
The new UI works, but I am still trying to make it pass the swagger validation, see the status here: http://online.swagger.io/validator/debug?url=https://raw.githubusercontent.com/thehyve/transmart-core/swagger-upgrade/open-api/swagger.json


gijskant:
Especially, `oneOf` and friends have been introduced in OpenAPI 3, so we need to upgrade to that standard. See, e.g., https://swagger.io/blog/news/whats-new-in-openapi-3-0/.

PR #347, opened 2018-09-11
Fix worker pool for parallel tasks to one per system.

gijskant:
The system would spawn a new pool of threads for each request/operation (e.g., per export or per counts request) instead of having one pool for the entire system. This caused problems with the number of database connections.

### DIFF
--- a/transmart-core-db/grails-app/services/org/transmartproject/db/support/SystemService.groovy
+++ b/transmart-core-db/grails-app/services/org/transmartproject/db/support/SystemService.groovy
@@ -80,13 +80,11 @@ class SystemService implements SystemResource {
     @Autowired
     SessionFactory sessionFactory
 
-
     RuntimeConfig getRuntimeConfig() {
         return modelMapper.map(runtimeConfig, RuntimeConfigRepresentation.class)
     }
 
     RuntimeConfig updateRuntimeConfig(@Valid RuntimeConfig config) {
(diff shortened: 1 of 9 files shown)


transmart-core-db/grails-app/services/org/transmartproject/db/support/SystemService.groovy
```diff
@@ -2,6 +2,7 @@ package org.transmartproject.db.support
 
 import grails.util.Holders
 import groovy.transform.CompileStatic
+import jsr166y.ForkJoinPool
```
ewelinagr: not used

gijskant:
I will move the `SystemConfig` bean from `transmart-api-server` to `transmart-core-db`.

gijskant:
Locally I got a weird error in `org.transmartproject.rest.v2.StorageControllerSpec`:
```
storageSystemPostTest

Condition not satisfied:

objectResponse.statusCode == OK
|              |          |  |
|              404        |  200
|                         false
<404 Not Found,Byte array resource [resource loaded from byte array],{Content-Type=[text/html;charset=utf-8], Content-Language=[en], Content-Length=[990], Date=[Tue, 11 Sep 2018 10:54:49 GMT]}>

	at org.transmartproject.rest.v2.StorageControllerSpec.storageSystemPostTest(StorageControllerSpec.groovy:138)
```

PR #350, opened 2018-09-12
Fix export data view selection and validation.

gijskant:


transmart-rest-api/grails-app/controllers/org/transmartproject/rest/ExportController.groovy
```diff
@@ -205,7 +206,11 @@ class ExportController {
      */
     def fileFormats(@RequestParam('dataView') String dataView) {
         checkForUnsupportedParams(params, ['dataView'])
-        def fileFormats = restExportService.getSupportedFormats(dataView)
+        def view = DataView.from(dataView)
```
ewelinagr: if the data view is case-insensitive, maybe it should be also changed in [line 88](https://github.com/thehyve/transmart-core/blob/bd5471608a38dd2bf1ac743a81ead1eeddde0301/transmart-rest-api/grails-app/controllers/org/transmartproject/rest/ExportController.groovy#L88)

gijskant: Good catch. Would have been picked up with static compilation, but grails...

gijskant:
I feel that the existing export functionality tries to do too much without doing it well. This TSV hypercube export is not really a useful export and the mix of data type, data view, etc, was not properly checked.
I could find noone that understood the highdim export of why is works as it works. Better to remove it now until we have time to build a proper highdim export.

PR #351, opened 2018-09-12
Fix access rights for job management.

gijskant:


gijskant:
- Access right checking was missing in some places
- Added `@CompileStatic` to the `ExportAsyncJobService`
- Fixed a JSON serialisation issue, cause by #350 (me).

transmart-rest-api/grails-app/services/org/transmartproject/rest/dataExport/ExportAsyncJobService.groovy
```diff
@@ -12,68 +18,85 @@ import org.transmartproject.db.multidimquery.query.InvalidQueryException
...
             newJob.jobName = newJob.id.toString()
             newJob.save()
         }
 
-        log.debug("Sending ${newJob.jobName} back to the client")
+        // Check for uniqueness of the job name for the user
+        def jobCount = DetachedCriteria.forClass(AsyncJobCoreDb)
+                .add(Restrictions.eq('jobName', newJob.jobName))
+                .add(Restrictions.eq('userId', user.username))
+                .getExecutableCriteria(sessionFactory.currentSession)
```
ewelinagr: so just to make sure, this will always find the new job that is saved in the session only (not flushed yet) and, if any, some result in the database? There should not be any gorm issues with mixing sessions or sth?

gijskant: Good question. My reasoning here is that this happens within a transaction (because of the `@Transactional` annotation), so within the same transaction it tries to save and check if the resulting number of jobs does not violate uniqueness. If that fails, it rolls back the transaction (reverting the save). For multiple simulatenous requests (with different sessions perhaps), I rely on the database to handle transactions correctly. Hopefully that works as intended...

PR #353, opened 2018-09-13
Update installation and upgrade documentation

ewelinagr:


gijskant:
Some suggestions:
- remove docs/README.md
- rename docs/installation.md to docs/README.md and refer from there to the upgrade manual for who needs to upgrade existing installations.
- Try to avoid 'click here' or 'this link' in documentation text, see, e.g., [advice on link texts from W3](https://www.w3.org/QA/Tips/noClickHere).

ewelinagr:
@gijskant thanks for the suggestions! I added some changes following your ideas. Please let me know, if there is something else we could improve.

PR #359, opened 2018-10-01
Add username to access error log message

forus:
Error message like "Access denied to concept code: 2cba26058f55a7d579f4a0a1f0b9a2e8d5e149a0" does not say which actually user does not have the access.


forus:
@ewelinagr good point. I don't think they diverged purposely.  I am going to add "or it does not exist" part to exception messages for concepts as well.

forus:
@ewelinagr I've changed the messages

PR #360, opened 2018-10-02
Fix hypercube serialisation for concurrent requests.

gijskant:
Create a new serialiser for each request.


gijskant:
Yes, but what was left was `interface HypercubeSerializer { void write() }`...

PR #361, opened 2018-10-03
Possibly fix an issue that observations calls fail under high load.

gijskant:


transmart-rest-api/grails-app/controllers/org/transmartproject/rest/QueryController.groovy
```diff
@@ -131,10 +131,8 @@ class QueryController extends AbstractQueryController {
         try {
             def args = new DataRetrievalParameters(constraint: constraint, sort: sort)
             hypercubeDataSerializationService.writeClinical(format, args, authContext.user, out)
-        } catch(LegacyStudyException e) {
+        } catch (LegacyStudyException e) {
             throw new InvalidRequestException("This endpoint does not support legacy studies.", e)
-        } finally {
-            out.close()
```
ewelinagr: Maybe similar fix is needed also for highdim? Line [467](https://github.com/thehyve/transmart-core/blob/116966d60ebf28ebf68bb290a5b472b55dacaf98/transmart-rest-api/grails-app/controllers/org/transmartproject/rest/QueryController.groovy#L467)

gijskant: Nice catch. Fixed it, let's see what the tests say...

PR #362, opened 2018-10-05
Fix data table serialisation

gijskant:
Write only dimension elements of current page of data table.
Includes #357.


forus:
@gijskant this PR contains two others?

PR #364, opened 2018-10-16
Use ISO-8601 to format observation date value

forus:


transmart-rest-api/src/main/groovy/org/transmartproject/rest/serialization/HypercubeProtobufSerializer.groovy
```diff
@@ -95,8 +96,10 @@ class HypercubeProtobufSerializer {
         if (value.value != null) {
             if (value.value instanceof Number) {
                 builder.numericValue = ((Number) value.value).doubleValue()
+            } else if (value.value instanceof Date) {
+                builder.stringValue = Instant.ofEpochMilli(((Date) value.value).time).toString()
```
ewelinagr: I am not sure what timestampValue (used below) is, so probably better to have it as a string

forus: protobuf does not support date format. People use an integer type for that.
Anyway, date type has been rendered as a string before. I only have changed format here.

PR #365, opened 2018-10-19
Don't rollback query set diffs when one has failed

forus:


transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
```diff
@@ -195,6 +196,7 @@ class UserQuerySetService implements UserQuerySetResource {
                 .list() as List<QuerySetInstance>
     }
 
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
```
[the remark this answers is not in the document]
ewelinagr: actually this will probably cause exact same issue described here: https://stackoverflow.com/a/28746991

transmart-core-db/grails-app/services/org/transmartproject/db/userqueries/UserQuerySetService.groovy
```diff
@@ -195,14 +188,22 @@ class UserQuerySetService implements UserQuerySetResource {
...
+        def queryRepresentation = UserQueryService.toRepresentation(query)
+        List<Long> newPatientIds = getPatientsForQuery(queryRepresentation, user)
 
         List<Long> addedIds = newPatientIds - previousPatientIds
         List<Long> removedIds = previousPatientIds - newPatientIds
 
         if (addedIds.size() > 0 || removedIds.size() > 0) {
             QuerySet querySet = new QuerySet()
-            querySet.query = query
+            querySet.query = Query.load(queryId)
```
ewelinagr: why not `query` object directly?

forus: `query` has `UserQuery` type here. We could choose to downcast like so `(Query) query`. But this won't work for a possible alternative implementation of `UserQuery` that is not a subclass of `Query`. e.g. Test mocks

 BTW `load` does not fetch the object from the database. It creates a proxy with id in it.

forus: hm, or we could simply fix the signature of the method to receive `Query` object. It's private one anyhow.

PR #374, opened 2018-11-14
Refresh materialzed view after migrating

forus:
Without doint so application would fail
until the after data loaing call is done


gijskant:
I don't understand this change. The view was just dropped and recreated, right?

forus:
@gijskant you are right. This change does not bring any value. I will close the PR.

PR #375, opened 2018-11-14
Fix data table with visit dimension serialization

ewelinagr:


transmart-core-db/src/main/groovy/org/transmartproject/db/multidimquery/DimensionImpl.groovy
```diff
@@ -740,8 +740,9 @@ class VisitDimension extends DimensionImpl<I2b2VisitDimension, VisitKey> impleme
     }
 
     @CompileDynamic
-    @Override List<I2b2VisitDimension> doResolveElements(List<VisitKey> elementKeys) {
-        elementKeys.collect { VisitKey key ->
+    @Override List<I2b2VisitDimension> doResolveElements(List elementKeys) {
```
gijskant: This looks fishy to me. It seems as some implementation of the hypercube incorrectly uses representation values instead of actual element keys (in this case the "encounter/patient" string instead of VisitKey objects). We should fix the underlying implementation instead of 'fixing' it here.

gijskant: The problem lies on line 715 and further, where a string is used as key. The comments mention that it "must be a number, string or date (used in e.g. json serialization) so VisitKey is not going to work here."
This is not true, the key can be serialised to string there, in the serialiser, instead of here.

ewelinagr:
@gijskant it's updated now

PR #376, opened 2018-11-14
Check access on patient set size

forus:
Users with count with threshold level of access
should not see original size of the patient set


transmart-core-db/grails-app/services/org/transmartproject/db/clinical/AggregateDataResourceImplService.groovy
```diff
@@ -47,18 +47,8 @@ class AggregateDataResourceImplService implements AggregateDataResource {
         User exactCountsAccessUserCopy = copyUserWithChangedPatientDataAccessLevel(user, COUNTS_WITH_THRESHOLD, COUNTS)
```
ewelinagr: now the user is copied twice - here and inside hidePatientCount, but it's minor

forus: yes, indeed. That's the price to pay to reuse the peace of code.

PR #380, opened 2019-01-03
Stop sharing security context between http sessions

forus:
Fixes TMT-658
When the security context is shared the authentification field gets mutated by multiple concurrent requests causing all sort of issues. Each request starts with setting authentification field with unauthenticated bean first and only after analyzing the token with authenticated.

The PR makes each request to use own security context instance.

The http session instance still could be shared between http requests when jsessionid is supplied in the cookie header.
TMT-664 is a ticket to asses how harmless it that.


forus:
@ewelinagr @gijskant FYI I've made two not related fixes in this PR. Please review 
fba991d and bbc788c

PR #381, opened 2019-01-09
Make ts copy support sparse ids

forus:
To be able to load current test_data.

This PR removes the requirement to the source data to specify ids in accordance with row number (row - 2 when we count with the header).

SIDE CHANGES:

 - Use `long` instead of `integer` type to represent source data ids (aka index)

TODO:
 - add a check that neither source id (aka index), nor result db id are not null. Add this check on lookup and before setting a new value.
 -  add check for duplicates. Smth like:
```
if (indexToPatientNum.containsKey(patientIndex)) {
	throw new IllegalStateException("Patient with patient_num=${patientIndex} has bee already inserted")
}
```

### DIFF
--- a/transmart-copy/src/main/groovy/org/transmartproject/copy/table/Dimensions.groovy
+++ b/transmart-copy/src/main/groovy/org/transmartproject/copy/table/Dimensions.groovy
@@ -30,7 +30,7 @@ class Dimensions {
     final LinkedHashMap<String, Class> columns
 
     final Map<String, Long> dimensionNameToId = [:]
-    final List<Long> indexToDimensionId = []
+    final Map<Long, Long> indexToDimensionId = [:]
     final Set<String> modifierDimensionCodes = []
 
     Dimensions(Database database) {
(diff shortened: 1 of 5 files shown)


gijskant:
Where does this change come from? It was explicitly designed to use subsequent numbers, e.g., to prevent mistakes, like missing data records.

gijskant:
Think about checking for duplicates.
Sorry, but I really don't see a use for sparse id's here. These numbers are supposed to be only an index. Let's not give people the impression that they can load whatever id's they like. Id's are generated by the system for you.

forus:
@gijskant Thank you for duplicates example.
checking for duplicates could and has to be done with the same amount of efforts for map-based approach as well of course.
```
if (indexToPatientNum.containsKey(patientIndex)) {
	throw new IllegalStateException("Patient with patient_num=${patientIndex} has already been inserted")
}
```
I've added it to TODO list of this PR.

Sorry, but I just don't understand what do we gain from having this constraint.

PR #382, opened 2019-01-09
Add CSR test study

forus:


transmart-data/test_studies/CSR/i2b2metadata/dimension_description.tsv
```diff
@@ -0,0 +1,7 @@
+id	density	modifier_code	value_type	name	packable	size_cd
+0				study		
+1				concept		
+2				patient		
+3	DENSE	CSR_DIAGNOSIS_MOD	T	Diagnosis ID	NOT_PACKABLE	SMALL
```
gijskant: I would name this dimension `Diagnosis` instead of `Diagnosis ID`

transmart-data/test_studies/Makefile
```diff
@@ -0,0 +1,15 @@
+SUBDIRS := $(wildcard */.)
```
gijskant: Odd choice to start using make scripts here again. Why would we choose make, and why is this part of transmart-data?

transmart-data/test_studies/Makefile
```diff
@@ -0,0 +1,15 @@
...
+TRANSMART_COPY_VERSION := 17.1-HYVE-5.7
+TRANSMART_COPY := transmart-copy-$(TRANSMART_COPY_VERSION).jar
+TRANSMART_COPY_SRC_URL := https://repo.thehyve.nl/content/repositories/releases/org/transmartproject/transmart-copy/$(TRANSMART_COPY_VERSION)/transmart-copy-$(TRANSMART_COPY_VERSION).jar
+
+$(TRANSMART_COPY):
+	curl -f $(TRANSMART_COPY_SRC_URL) > $@
+
+load: $(SUBDIRS)
+$(SUBDIRS): $(TRANSMART_COPY)
+	@echo "Laading test study from $@ folder."
```
gijskant: Typo (Laading -> Loading)

transmart-data/test_studies/Makefile
```diff
@@ -0,0 +1,15 @@
+SUBDIRS := $(wildcard */.)
+
+TRANSMART_COPY_VERSION := 17.1-HYVE-5.7
+TRANSMART_COPY := transmart-copy-$(TRANSMART_COPY_VERSION).jar
+TRANSMART_COPY_SRC_URL := https://repo.thehyve.nl/content/repositories/releases/org/transmartproject/transmart-copy/$(TRANSMART_COPY_VERSION)/transmart-copy-$(TRANSMART_COPY_VERSION).jar
```
gijskant: Well, the source is located at `../../transmart-copy`...

transmart-data/test_studies/CSR/i2b2metadata/dimension_description.tsv
forus: I see. We have "<ORG> Diagnosis ID" in the customer DB. I guess it's named that way to stress that it contains diagnosis per patient (Ids) as opposed to actual diagnosis names.
My consideration here was to have a dimension name that contains space.

transmart-data/test_studies/Makefile
forus: true, but we release `transmart-data` as a separate component as well.

In dev setup, you could always build `transmart-copy` from sources and copy to the folder. `make load` won't fetch remote one if the version is right (`TRANSMART_COPY_VERSION `)

transmart-data/test_studies/Makefile
forus: Make does not bring much here indeed. `bash` would do as well.
The rest of test studies that are used by automatic tests are part of `transmart-data` currently.
So you could create the test ready db with `make postgres_test`.
Please feel free to create a ticket for separating test data from `transmart-data` where would decide on a new approach.

PR #385, opened 2019-01-31
Add dimensions endpoint; refactor dimensions serialisation.

gijskant:
Fixes [TMT-670](https://jira.thehyve.nl/browse/TMT-670)


forus:
IMHO it's hard to review code that has refactoring and func. changes in one commit. I'd prefer to have them in separate commits. Improving the open-api documentation looks like a perfect candidate for a separate commit.

It seems to be tricky at this point to split the changes to separate commits. So perhaps something for the future.

transmart-rest-api/grails-app/controllers/org/transmartproject/rest/DimensionController.groovy
```diff
@@ -70,12 +61,87 @@ class DimensionController extends AbstractQueryController {
...
+        response.contentType = 'application/json'
+        response.characterEncoding = 'utf-8'
+        new ObjectMapper().writeValue(response.outputStream, [dimensions: dimensions])
+    }
+
+    /**
+     * Dimension endpoint:
+     * <code>/v2/dimensions/${dimensionName}</code>
+     *
+     * @return the dimension properties if the user has access to it.
```
forus: I'd expect from this call to return single dimension object instead.
Should this call have `/v2/dimensions/${dimensionName}/properties` path instead?

gijskant: That is exactly what is gives, a single dimension object instead. But not a dimension as in all elements of one dimension of a hypercube (something I don't like about the current Dimension interface: it combines dimension as an object with retrieval of dimension elements), but just its properties.

transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/MultiDimensionalDataResource.groovy
```diff
@@ -44,7 +44,7 @@ interface MultiDimensionalDataResource {
      * @param user
      * @param constraint
      */
-    Iterable getDimensionElements(Dimension dimension, Constraint constraint, User user)
```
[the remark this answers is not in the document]
gijskant: In my opinion a cleaner and lighter interface. My intention was that the `listElements` method in the dimension controller would not need to fetch and pass a Dimension object, but I could not get around passing a dimension object to the serialiser. In my opinion too much logic is in these dimension objects: not just names and types, but also serialisation.

PR #387, opened 2019-02-06
Fix variable metadata classes and documentation.

gijskant:


open-api/swagger.yaml
```diff
@@ -3211,17 +3216,20 @@ components:
             type: string
         metadata:
           $ref: '#/components/schemas/StudyMetadata'
+
     StudyMetadata:
       type: object
       properties:
-        conceptToVariableName:
+        conceptCodeToVariableMetadata:
           description: a map from concept code to variable name in the study.
```
ewelinagr: Looks like the description could be updated as well to 
> a map from concept code to variable metadata...

PR #389, opened 2019-02-06
Add dimension description columns

gijskant:


gijskant:
It's `i2b2metadata` now.

transmart-copy/src/main/groovy/org/transmartproject/copy/table/TreeNodes.groovy
```diff
@@ -137,7 +138,7 @@ class TreeNodes {
             log.info "Reading tree nodes from file ..."
             def insertCount = 0
             def existingCount = 0
-            LinkedHashMap<String, Class> header = columns
+            LinkedHashMap<String, Class> header
```
forus: Why do you have the variable declaration without initialization? There are several similar declarations above.

forus:
There is no test coverage for fields retrieval. Sorry, I've missed it the first time.

forus:
It's nice to have this change in ts-copy 👍

transmart-copy/src/main/groovy/org/transmartproject/copy/table/TreeNodes.groovy
gijskant: Because of the changed type. This initialisation that was there was actually not used.

PR #403, opened 2019-03-29
Support the visit dimension in transmart-copy.

gijskant:


transmart-copy/src/test/resources/examples/SURVEY/i2b2demodata/observation_fact.tsv
```diff
@@ -1,14 +0,0 @@
-encounter_num	patient_num	concept_cd	provider_id	start_date	modifier_cd	instance_num	trial_visit_num	valtype_cd	tval_char	nval_num	valueflag_cd	quantity_num	units_cd	end_date	location_cd	observation_blob	confidence_num	update_date	download_date	import_date	sourcesystem_cd	upload_id	sample_cd
```
ewelinagr: why this file needs to be removed?

gijskant: Oops, will restore it.

PR #408, opened 2019-04-24
Docker: Add Dockerfile for transmart-api-server

lukaszimmermann:
Adds the initial version of the `Dockerfile` for the transmart-api-server.

Some notes:
-  The `Dockerfile` could also have a `HEALTHCHECK` instruction.
-  Other configuration items can be exposed later via environment variables on demand.
- `docker-compose.yml` file was updated to give an example on the runtime information of that image.
- Logging is on `stdout` and available via the `logs` command of `docker` and `docker-compose` CLI. The log level is not yet configurable.
- Image not yet available on Docker Hub. Need to check required permissions with The Hyve.


docker/transmart-api-server/docker-entrypoint.sh
```diff
@@ -0,0 +1,142 @@
...
+EndOfMessage
+		exit 1
+}
+
+# Check that Postgres and Keycloak Host are configured
+# via environment variables
+[ ! -z ${PGHOST+x} ] || fatal 'PGHOST'
+[ ! -z ${PGPORT+x} ] || fatal 'PGPORT'
+[ ! -z ${KEYCLOAK_HOST+x} ] || fatal 'KEYCLOAK_HOST'
+[ ! -z ${KEYCLOAK_PORT+x} ] || fatal 'KEYCLOAK_PORT'
```
ewelinagr: having one `KEYCLOAK_URL` variable instead of separate ones for host and port seems to be a better option - less variables plus enables referring to keycloak via DNS.
It might be useful to add `KEYCLOAK_CLIENT` and `KEYCLOAK_REALM` variables too, but this is something we could discuss later.

docker/transmart-api-server/docker-entrypoint.sh
```diff
@@ -0,0 +1,142 @@
...
+
+# Fixed values, not configurable by user
+PGDATABASE=transmart
+BIOMART_USER='biomart_user'
+BIOMART_PASSWORD="${BIOMART_USER}"
+TRANSMART_API_SERVER_CONFIG_FILE="${TRANSMART_USER_HOME}/transmart-api-server.config.yml"
+
+cat > "${TRANSMART_API_SERVER_CONFIG_FILE}" <<EndOfMessage
```
ewelinagr: The content of this config file could be significantly reduced. Most of the properties are set by default based on the internal application.yml file that is packaged inside the application jar. 
The properties specified here will override the ones from internal (inside-jar) config, so the file generated here could be limited to something like this:

````
---
dataSource:
    driverClassName: org.postgresql.Driver
    dialect: org.hibernate.dialect.PostgreSQLDialect
    url: jdbc:postgresql://${PGHOST}:${PGPORT}/${PGDATABASE}
    username: ${PGUSER}
    password: ${PGPASSWORD}
    dbCreate: none
    pooled: true
    jmxExport: true
    logSql: false
    formatql: false
    properties:
        jmxEnabled: true
        initialSize: 5
        maxActive: 50
        minIdle: 5
        maxIdle: 25
        maxWait: 10000
        maxAge: 600000
        timeBetweenEvictionRunsMillis: 5000
        minEvictableIdleTimeMillis: 60000
        validationQuery: select 1
        validationQueryTimeout: 3
        validationInterval: 15000
        testOnBorrow: true
        testWhileIdle: true
        testOnReturn: false
        jdbcInterceptors: ConnectionState
        defaultTransactionIsolation: 2 # TRANSACTION_READ_COMMITTED

keycloak:
    realm: ${KEYCLOAK_REALM}
    bearer-only: true
    auth-server-url: ${KEYCLOAK_URL}
    resource: ${KEYCLOAK_CLIENT}
    use-resource-role-mappings: true

keycloakOffline:
    offlineToken: <offline_token>

org.transmartproject.patientCountThreshold: 5

```

docker/docker-compose.yml
```diff
@@ -1,4 +1,3 @@
----
 version: '2.0'
```
ewelinagr: could it be upgraded to version 3?

docker/docker-compose.yml
```diff
@@ -8,3 +7,15 @@ services:
         build: ./solr
```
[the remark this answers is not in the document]
lukaszimmermann: I thought that this `docker-compose.yml` was intended to deploy all of tranSMART, and not only the API server. The file would then also be misplaced here.

Also I did not author this file, but @PiotrZakrzewski  back in 2016.

docker/docker-compose.yml
lukaszimmermann: Sure, I did not set this version

lukaszimmermann:
I think we can now merge this and then discuss how Glowing Bear would then also be included (either in this `docker-compose.yml` file or in a new one in another repository).

PR #410, opened 2019-05-01
TranSMART database creation using liquibase.

gijskant:


transmart-core-db/grails-app/services/org/transmartproject/db/log/AccessLogService.groovy
```diff
@@ -16,6 +17,10 @@ class AccessLogService implements AccessLogEntryResource {
 
     static final Logger log = LoggerFactory.getLogger(AccessLogService.class)
 
+    boolean writeLogToDatabase() {
+        Holders.config.getProperty('org.transmartproject.system.writeLogToDatabase', Boolean.class, true)
```
ewelinagr: Maybe it would be useful to document this one somewhere?

transmart-schemas/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+### How to autogenerate a changelog
+
+- Download [Liquibase](https://download.liquibase.org/)
+- Download [postgresql-42.2.5.jre7.jar](https://search.maven.org/remotecontent?filepath=org/postgresql/postgresql/42.2.5.jre7/postgresql-42.2.5.jre7.jar)
+- Create a file `liquibase.properties`:
+    ```properties
+    driver=org.postgresql.Driver
+    classpath=postgresql-42.2.5.jre7.jar
+    url=jdbc:postgresql://localhost:5432/transmart
```
ewelinagr: Db name here is different then the one described in the [liquidbase.propertis](https://github.com/thehyve/transmart-core/pull/410/files#diff-178406dbbb43e959f0e8d82e15639647R3)

transmart-schemas/rebuild.sh
```diff
@@ -0,0 +1,35 @@
...
+if [[ ! "x${JAVA_HOME}" == "x" ]]; then
+    prefix="${JAVA_HOME}/bin/"
+fi
+
+
+sudo -u postgres psql -c 'drop database "transmartDev";'
+sudo -u postgres psql -c 'create database "transmartDev";'
+sudo -u postgres psql -c 'grant all on database "transmartDev" to tm_admin;'
+${prefix}java -jar -Dspring.config.location=./database.yml "${jar}"
+sudo -u postgres psql -d transmartDev -c 'grant all on schema i2b2demodata to biomart_user;'
```
ewelinagr: Shouldn't `biomart_user` and `tm_cz` be created first?

transmart-schemas/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+apply plugin: 'io.spring.dependency-management'
+
+ext {
+    liquibaseVersion = '3.6.1'
+}
+
+dependencies {
+    compile 'org.springframework.boot:spring-boot-starter'
+    compile "org.liquibase:liquibase-core:${liquibaseVersion}"
+    compile "org.postgresql:postgresql:${postgresqlVersion}"
```
ewelinagr: where should the `${postgresqlVersion}` be configured?

gijskant: That is configured in the root [build.gradle](build.gradle)

transmart-schemas/rebuild.sh
gijskant: Yes, but there is no `create role ... if not exists`... I will try some other way

PR #416, opened 2019-05-10
Escape invalid file name characters in export names

ewelinagr:
Fixes TMT-836

### DIFF
--- a/transmart-rest-api/grails-app/services/org/transmartproject/rest/dataExport/ExportAsyncJobService.groovy
+++ b/transmart-rest-api/grails-app/services/org/transmartproject/rest/dataExport/ExportAsyncJobService.groovy
@@ -56,12 +56,7 @@ class ExportAsyncJobService extends AbstractDataResourceService {
 
     AsyncJobCoreDb createNewJob(String jobName, User user) {
         // Check input
-        if (jobName != null) {
-            jobName = jobName.trim()
-            if (jobName.empty) {
-                jobName = null
-            }
-        }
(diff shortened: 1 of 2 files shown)


transmart-rest-api/grails-app/services/org/transmartproject/rest/dataExport/ExportAsyncJobService.groovy
```diff
@@ -160,6 +155,11 @@ class ExportAsyncJobService extends AbstractDataResourceService {
         new JobKey(jobId.toString(), 'DataExport')
     }
 
+    private static escapeInvalidFileNameChars(String fileName) {
+        // escape characters that are not on the file name characters whitelist
+        fileName?.replaceAll("[^a-zA-Z0-9_\\.\\-\\s]", "_")
```
gijskant: This quite restrictive, would also replace characters like ë, ø, ń with _. It could make the job names harder to read and increases the probability is job name collisions (we anyway need to adapt the [ExcludeValuesValidatorDirective](https://github.com/thehyve/glowing-bear/blob/dev/src/app/validators/exclude-values-validator.directive.ts#L18) in Glowing Bear to prevent duplicate job names).

An alterative could be to just replace the [reserved characters](https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words) (except `.` and ` `).

ewelinagr: @gijskant I updated the replacing rule and addressed the related Glowing Bear change in [TMT-839](https://jira.thehyve.nl/browse/TMT-839)

PR #417, opened 2019-05-14
Expect the same KEYCLOAK_SERVER_URL in all glowing-bear-docker components

ewelinagr:
Expect `https://keycloak-example.com` (without `/auth` part)


ewelinagr:
Change required in https://github.com/thehyve/glowing-bear-docker/blob/master/docker-compose.yml#L81 after merging.

PR #419, opened 2019-05-23
Enable incremental loading of patient data for a study

ewelinagr:
Implements [TMT-832](https://jira.thehyve.nl/browse/TMT-832)
- incremental study loading can be triggered by a new command line option: `-I` or `--incremental`
- with incremental option not all observations for the studies in the input data are deleted before loading (default behavior), but only the observations for the patients in the input data

Scenarios covered by tests:
- [x] inserts of new patient records in a next upload batch for a study, 
- [x] updates of existing patients,
- [x] deletes of patient records (present in a first upload part, but not in a second one),  
- [x] data for a patient is present in multiple studies

### DIFF
--- a/transmart-copy/src/main/groovy/org/transmartproject/copy/Copy.groovy
+++ b/transmart-copy/src/main/groovy/org/transmartproject/copy/Copy.groovy
@@ -101,19 +101,20 @@ class Copy implements AutoCloseable {
         dimensions.fetch()
         dimensions.load(rootPath)
 
+        patients = new Patients(database)
+        patients.fetch()
+        patients.load(rootPath)
+
         // Insert study, trial visit objects
         studies = new Studies(database, dimensions)
         if (incremental) {
             studies.fetch()
             studies.fetchStudyDimensionDescriptions()
+            deleteObservationsForPatientsInStudies(rootPath)
         } else {
             studies.check(rootPath)
         }
         studies.load(rootPath)
-
-        patients = new Patients(database)
-        patients.fetch()
-        patients.load(rootPath)
 
         def concepts = new Concepts(database, config.updateConceptPaths)
         concepts.fetch()
@@ -169,14 +170,10 @@ class Copy implements AutoCloseable {
 
     void deleteObservationsForPatientsInStudies(String rootPath) {
         log.info "Deleting observations for patients found in ${rootPath} ..."
-        def dimensions = new Dimensions(database)
-        studies = new Studies(database, dimensions)
         def studyIds = studies.readStudyIds(rootPath)
         for (String studyId : studyIds) {
             def trialVisitNums = studies.findTrialVisitNumsForStudyId(studyId)
-            patients = new Patients(database)
-            patients.fetch()
-            patients.removeObservations(rootPath, trialVisitNums as Set)
+            patients.removeObservations(trialVisitNums as Set)
         }
     }
 
@@ -247,13 +244,11 @@ class Copy implements AutoCloseable {
                     copy.uploadPedigree(directory, config)
                 }
                 if (!modes || 'study' in modes) {
-                    if (cl.hasOption('incremental')) {
-                        copy.deleteObservationsForPatientsInStudies(directory)
-                        copy.uploadStudy(directory, config, true)
-                    } else {
+                    boolean incremental = cl.hasOption('incremental')
+                    if (!incremental) {
                         copy.deleteStudy(directory, false)
-                        copy.uploadStudy(directory, config, false)
                     }
+                    copy.uploadStudy(directory, config, incremental)
                 }
             }
             if (cl.hasOption('restore-indexes')) {
--- a/transmart-copy/src/main/groovy/org/transmartproject/copy/Copy.groovy
+++ b/transmart-copy/src/main/groovy/org/transmartproject/copy/Copy.groovy
@@ -57,6 +57,8 @@ class Copy implements AutoCloseable {
                 'Specifies a data directory.')
         options.addOption('m', 'mode', true,
                 'Load mode (e.g. \'study\' or \'pedigree\').')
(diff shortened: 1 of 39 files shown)


transmart-copy/src/main/groovy/org/transmartproject/copy/Database.groovy
```diff
@@ -91,11 +91,18 @@ class Database implements AutoCloseable {
         if (!password) {
             throw new IllegalArgumentException('Please set the PGPASSWORD environment variable.')
         }
+        Integer maximumPoolSize
+        try{
+            maximumPoolSize = Integer.parseInt(params['MAXPOOLSIZE'])
+        } catch (NumberFormatException e) {
+            maximumPoolSize = 8
```
gijskant: I don't see the need for this parameter, what is the use case?
The default value in Hikari seems to be [10](https://github.com/brettwooldridge/HikariCP#configuration-knobs-baby), is there a reason to have 8 here?
About the code: I would expect something like
```groovy
if (params['MAXPOOLSIZE']) {
  config.maximumPoolSize = Integer.parseInt(params['MAXPOOLSIZE'])
}
```
i.e., only setting it when a parameter is provided, using the default of Hikari otherwise.
Also, I would expect an error to be thrown when it can't parse instead of silently using a default instead.

transmart-copy/src/main/groovy/org/transmartproject/copy/Copy.groovy
```diff
@@ -228,8 +247,13 @@ class Copy implements AutoCloseable {
                     copy.uploadPedigree(directory, config)
                 }
                 if (!modes || 'study' in modes) {
-                    copy.deleteStudy(directory, false)
-                    copy.uploadStudy(directory, config)
+                    if (cl.hasOption('incremental')) {
+                        copy.deleteObservationsForPatientsInStudies(directory)
```
gijskant: I would expect this to be part of `uploadStudy`. The idea is that all data is read only once, single pass, and that the table classes are instantiated only once.

transmart-copy/src/main/groovy/org/transmartproject/copy/Database.groovy
ewelinagr: I added it because with the default max pool size I was getting this error, when running tests:
`org.postgresql.util.PSQLException: FATAL: sorry, too many clients already`
I will change it the way you mentioned.

transmart-copy/src/main/groovy/org/transmartproject/copy/table/Patients.groovy
```diff
@@ -149,4 +149,38 @@ class Patients {
...
+            }
+        }
+    }
+
+    private void removeObservationsForPatientAndTrials(Long patientNum, Set<Long> trialVisitNums) {
+        if (!trialVisitNums) {
+            return
+        }
+        int observationCount = database.namedParameterJdbcTemplate.update(
+                """delete from ${Observations.TABLE} where
```
gijskant: This could be a single delete for all these patients, right? (but then in chunks of max_parameters)

gijskant: In the `void load(String rootPath)` function, the existing patients are already available.

transmart-copy/src/test/groovy/org/transmartproject/copy/CopySpec.groovy
```diff
@@ -283,6 +285,141 @@ class CopySpec extends Specification {
...
+        inDbTreeNodePaths2.containsAll(expectedTreeNodePaths2)
+        inDbSubjectIds2.findAll{ it.toString().startsWith("SURVEY_INC_")} as Set ==
+                (expectedSubjectIds1 + expectedSubjectIds2) as Set
+
+        // observation for patient from part1 only - old value
+        readFieldsFromDb(Observations.TABLE,'nval_num',
+                "where patient_num=${inDbPatientIdeToPatientNum['SURVEY_INC_P2']} AND concept_cd='age'")[0] == 33
+        // observation for patient updated in part2 - updated value
+        readFieldsFromDb(Observations.TABLE,'nval_num',
+                "where patient_num=${inDbPatientIdeToPatientNum['SURVEY_INC_P1']} AND concept_cd='age'")[0] == 26
```
gijskant: It may be more convenient to test if `readFieldsFromDb(...) == [26]` and it also tests the number of observations for the patient.

transmart-copy/src/test/resources/examples/SURVEY_INC/part1/i2b2metadata/i2b2_secure.tsv
```diff
@@ -0,0 +1,4 @@
+c_hlevel	c_fullname	c_name	c_synonym_cd	c_visualattributes	c_totalnum	c_basecode	c_metadataxml	c_facttablecolumn	c_tablename	c_columnname	c_columndatatype	c_operator	c_dimcode	c_comment	c_tooltip	m_applied_path	update_date	download_date	import_date	sourcesystem_cd	valuetype_cd	m_exclusion_cd	c_path	c_symbol	i2b2_id	secure_obj_token
+3	\Projects\Survey_Inc\part1\Demographics\Gender\	Gender	N	LAC		gender		CONCEPT_CD	CONCEPT_DIMENSION	CONCEPT_PATH	T	LIKE	\Demographics\Gender\			@				SURVEY_INC						SURVEY_INC
```
gijskant: This works for the tests, but it wouldn't show up in the tree in TranSMART due to missing parent nodes if we would like to use this test data set for other purposes.

ewelinagr:
Thanks for the review, @gijskant ! I added the improvements here: https://github.com/thehyve/transmart-core/pull/419/commits/e9ac8c6ce1b641444972a418d18e269f3c36f32a

PR #423, opened 2019-06-13
Update documentation of transmart-copy parameters.

gijskant:


transmart-copy/src/main/groovy/org/transmartproject/copy/Copy.groovy
```diff
@@ -238,12 +238,13 @@ class Copy implements AutoCloseable {
                         unlogged: cl.hasOption('unlogged'),
                         baseOnMaxInstanceNum: cl.hasOption('base-on-max-instance-num')
                 )
-                def modes = cl.getOptionValues('mode')
+                // Read modes, use 'study' as default
+                String[] modes = cl.getOptionValues('mode') ?: ['study'] as String[]
```
gijskant: Interestingly this does not work, while IDE and compiler won't complain:
```groovy
def modes = cl.getOptionValues('mode') ?: ['study']
```
Also not runtime errors. I wonder how groovy does interpret the `'study' in modes` then...

ewelinagr: good to be aware of this!

PR #425, opened 2019-06-20
Encounter mapping

gijskant:


transmart-core-db/src/integration-test/groovy/org/transmartproject/db/multidimquery/QueryServicePgSpec.groovy
```diff
@@ -463,13 +463,9 @@ class QueryServicePgSpec extends Specification {
...
+        then: "List of all visits matching the constraints is returned"
         visits.size() == expectedResult.size()
-        expectedResult.every { expect ->
-            visits.any {
-                [expect.id] == [it.id]
-            }
-        }
+        visits.collect { it.encounterIds['VISIT_ID'] }.sort() == expectedResult.collect { it.encounterIds['VISIT_ID']}.sort()
 
         //TODO Fix counting elements of visit dimension - find a way to count distinct on two properties
```
ewelinagr: isn't this one fixed now?

gijskant: I should indeed have been fixed now, having a single identifier. But it seems that `getDimensionElementsCount` does not exist anymore. I will remove these lines from the test.

transmart-data/test_studies/Makefile
```diff
@@ -1,6 +1,6 @@
 SUBDIRS := $(wildcard */.)
 
-TRANSMART_COPY_VERSION := 17.1-HYVE-5.7
+TRANSMART_COPY_VERSION := $(shell cat ../VERSION)
```
ewelinagr: should this file be added then? Or is it auto-generated?

gijskant: It is auto-generated here: [transmart-data/build.gradle](https://github.com/thehyve/transmart-core/blob/dev/transmart-data/build.gradle#L152).
This file is always included in distributions (the transmart-data-<version>.zip releases) and is also a dependency for the setupPostgresTest task.

PR #426, opened 2019-06-25
Update visit dimension database documentation.

gijskant:
Fixes [TMT-918](https://jira.thehyve.nl/browse/TMT-918).


README.md
```diff
@@ -20,10 +22,13 @@ The OAuth2 authentication of the API for transmart-server is managed by [transma
 
 As an user interface for transmart-server there is a frontend application [transmartApp](transmartApp) built as a "web" profiled Grails platform plugin. There is also a modern cohort selector for i2b2 tranSMART named [Glowing Bear](https://github.com/thehyve/glowing-bear) that can used both with transmart-server and transmart-api-server.
 
+The database schema is described in the [data model documentation](docs/data-model.md).
+The schema is created using Liquibase at application startup, when it is configured to do so
+(see [Setup database](docs#2-setup-database)).
+The [transmart-copy](transmart-copy) tool can be used to load data into the database.
 
-Database definitions and installation instructions are in [transmart-data](transmart-data).
+Legacy database definitions and installation instructions are in [transmart-data](transmart-data).
```
ewelinagr: In the [installation instruction](https://github.com/thehyve/transmart-core/tree/visit-dimension-docs/docs#2-setup-database) this is still described as a main way of setting up the database. On the other hand the installation instruction is in general a bit outdated (still describing the transmart-server as the main app), so updating it can be done as a separate ticket, including also a link to docker installation instruction.

README.md
```diff
@@ -20,10 +22,13 @@ The OAuth2 authentication of the API for transmart-server is managed by [transma
 
 As an user interface for transmart-server there is a frontend application [transmartApp](transmartApp) built as a "web" profiled Grails platform plugin. There is also a modern cohort selector for i2b2 tranSMART named [Glowing Bear](https://github.com/thehyve/glowing-bear) that can used both with transmart-server and transmart-api-server.
 
+The database schema is described in the [data model documentation](docs/data-model.md).
+The schema is created using Liquibase at application startup, when it is configured to do so
+(see [Setup database](docs#2-setup-database)).
```
ewelinagr: Maybe also usefull to mention that both ways should work for postgres and oracle (not sure if the liquibase works with oracle though, it should, right?)

docs/data-model.md
```diff
@@ -0,0 +1,33 @@
+# TranSMART data model
+
+## TranSMART database model
+
+This diagram gives an overview of the core tables in the TranSMART database.
+The central table stores observations, which are linked to patients, concepts,
+visits, trial visits and studies.
+
+![TranSMART database diagram](model/transmart-database/transmart%20database%20diagram.jpg)
```
ewelinagr: The field names marked with **bold** text are not clear for me. Were they supposed to be PK/FK only or all not-nullable fields? It seems they are a bit inconsistent.

README.md
gijskant: Yes indeed, and hopefully together with a cleaned up repository (without legacy components).

README.md
gijskant: I created and tested it for `postgresql` and `h2`. I doubt if it works for Oracle, because of the `serial` data type. Should not be hard to make it work, but we would need time to test. I will add a note about it.

docs/data-model.md
gijskant: The intention is that those in bold are the non-nullable fields.

ewelinagr: hmm, so it looks like some of the fields are missed, eg. observation_fact.provider_id, observation_fact.modifier_cd and e.g. concept_code.name_char is in bold, but shouldn't.

gijskant: provider_id and modifier_cd are pseudo-nullable, i.e., '@' means no provider or no modifier. I will change the name field in the concept dimension.

gijskant: It's a bit weird, they should not be part of the primary key, in my opinion. Then they can be properly nullable and have foreign key constraints.

PR #432, opened 2019-07-04
Enable importing certificates into the Java keystore.

gijskant:


docker/transmart-api-server/docker-entrypoint.sh
```diff
@@ -52,6 +54,10 @@ keycloak:
 EndOfMessage
 sync
 
+[[ -e "${CERTS_PATH}" ]] && \
```
ewelinagr: maybe better `-f` (to make sure it's file, not dir)?

gijskant: Yes, I'll change it.

PR #440, opened 2019-07-10
Add relations endpoint

ewelinagr:
Related ticket: [TMT-934](https://jira.thehyve.nl/browse/TMT-934)

### DIFF
--- a/transmart-core-db/grails-app/services/org/transmartproject/db/ontology/MDStudiesService.groovy
+++ b/transmart-core-db/grails-app/services/org/transmartproject/db/ontology/MDStudiesService.groovy
@@ -26,6 +26,8 @@ import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentSkipListMap
 import java.util.stream.Collectors
 
+import static org.transmartproject.core.users.AuthorisationHelper.getPUBLIC_TOKENS
+
 @Transactional
 @CompileStatic
 class MDStudiesService implements MDStudiesResource, ApplicationRunner {
@@ -153,6 +155,15 @@ class MDStudiesService implements MDStudiesResource, ApplicationRunner {
             throw new NoSuchResourceException("No study found for specified studyIds.")
         }
         studiesForUser
+    }
+
+    Boolean areAllStudiesPublic() {
+        def criteria = DetachedCriteria.forClass(Study)
+        criteria.add(Restrictions.not(Restrictions.in('secureObjectToken', PUBLIC_TOKENS)))
+        List<MDStudy> privateStudies = criteria.getExecutableCriteria(sessionFactory.currentSession)
+                .list().stream()
+                .collect(Collectors.toList())
+        privateStudies.size() == 0
     }
 
     private void checkAccessToStudy(Study study, User user, id) {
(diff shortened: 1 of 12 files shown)


transmart-core-db/grails-app/services/org/transmartproject/db/ontology/MDStudiesService.groovy
```diff
@@ -155,6 +157,15 @@ class MDStudiesService implements MDStudiesResource, ApplicationRunner {
         studiesForUser
     }
 
+    Boolean areAllStudiesPublic() {
+        def criteria = DetachedCriteria.forClass(Study)
+        criteria.add(Restrictions.not(Restrictions.in('secureObjectToken', PUBLIC_TOKENS)))
+        List<MDStudy> privateStudies = criteria.getExecutableCriteria(sessionFactory.currentSession)
+                .list().stream()
```
gijskant: count instead of list?

transmart-core-db/grails-app/services/org/transmartproject/db/pedigree/RelationService.groovy
```diff
@@ -0,0 +1,27 @@
...
+@Transactional(readOnly = true)
+@CompileStatic
+class RelationService implements RelationResource {
+
+    @Autowired
+    MDStudiesService studiesService
+
+    @Override
+    List<org.transmartproject.core.pedigree.Relation> getAll(User user) throws AccessDeniedException {
+        if (studiesService.areAllStudiesPublic() || user.admin) {
```
gijskant: Just as a general remark:
There are generally two approaches in this kind of functions:
```java
if (errorCondition) {
  throw Exception()
}
doSomething()
```
or
```java
if (!errorCondition) {
  doSomething()
} else {
  throw Exception()
}
```

I personally like the first one more, because the error condition and the error response are together, and the indenting signals that the main behaviour (`doSomething()`) is part of the main flow, and not just some branch of behaviour. Do you agree?
It's a matter of code style, so I don't think there is only one good approach.

transmart-core-db/grails-app/services/org/transmartproject/db/pedigree/RelationService.groovy
```diff
@@ -0,0 +1,27 @@
...
+
+@Transactional(readOnly = true)
+@CompileStatic
+class RelationService implements RelationResource {
+
+    @Autowired
+    MDStudiesService studiesService
+
+    @Override
+    List<org.transmartproject.core.pedigree.Relation> getAll(User user) throws AccessDeniedException {
```
gijskant: In TranSMART things are typically done like this: service methods return domain object (Hibernate wrapper objects), but referred to by some interface. But still all the session stuff and lazy evaluation magic is still there.
What we do in other project (e.g., whereabouts) could be applied here to: stream to RelationRepresentation objects here (preferably using Java stream() instead of groovy (dynamic) collect).
I know that we deviate from other places in TranSMART if we start doing that, but I do think it is a change in the right direction.

transmart-core-db/grails-app/services/org/transmartproject/db/pedigree/RelationService.groovy
ewelinagr: yes, indeed the first approach looks better. In this case I used the second one, because this condition: `if(c1 || c2)` seemed simpler to me than `if(!c1 && !c2)`, but maybe better to be consistent and use the first one in all the cases. I updated the code here.

PR #445, opened 2019-07-15
Skip pg_bitcount schema updates until extension is available.

gijskant:


transmart-schemas/src/main/resources/db/changelog/changes/20191907041159_biomart_user_views.xml
```diff
@@ -5,16 +5,15 @@
   ~  (see accompanying file LICENSE).
   -->
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
-                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
```
phozzy: Don't we need ext (extension?) schemas?

phozzy: https://liquibase.jira.com/wiki/spaces/CONTRIB/overview - looks like we don't

PR #448, opened 2019-07-24
Fix invalid constraint rewrite rule.

gijskant:
The rule is not valid when the subselection dimension is
different from its parent dimension.

### DIFF
--- a/transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/query/NormaliseConstraintRewriter.groovy
+++ b/transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/query/NormaliseConstraintRewriter.groovy
@@ -15,12 +15,12 @@ import groovy.transform.CompileStatic
 class NormaliseConstraintRewriter extends ConstraintRewriter {
 
     /**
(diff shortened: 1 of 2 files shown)


transmart-core-api/src/main/groovy/org/transmartproject/core/multidimquery/query/NormaliseConstraintRewriter.groovy
```diff
@@ -14,19 +14,6 @@ import groovy.transform.CompileStatic
 @CompileStatic
 class NormaliseConstraintRewriter extends ConstraintRewriter {
 
-    /**
-     * Eliminate subselect true
```
ewelinagr: Maybe still useful to keep this for patient dimension only?

gijskant: You're right! I will change it.

gijskant: Fixed

PR #450, opened 2019-07-29
Include Health Check Indicators for application status monitoring

ewelinagr:
Add `/health` endpoint to monitor availability of the application.

Related ticket: [TMT-944]( https://jira.thehyve.nl/browse/TMT-944)

### DIFF
--- a/open-api/swagger.yaml
+++ b/open-api/swagger.yaml
@@ -13,13 +13,29 @@ info:
 
     # OAuth2
 
-    All calls need an [Authorization header](https://wiki.transmartfoundation.org/display/transmartwiki/tranSMART+REST+API).
+    All tranSMART REST API calls need an [Authorization header](https://wiki.transmartfoundation.org/display/transmartwiki/tranSMART+REST+API).
 
     ```
 
     Authorization:Bearer {token}
 
     ```
+
+
+    # Monitoring
+
+    For the application availability monitoring there is a
+    [Spring Boot Actuator health endpoint](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html) configured.
+
+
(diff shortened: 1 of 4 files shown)


open-api/swagger.yaml
```diff
@@ -588,6 +588,20 @@ paths:
                 $ref: '#/components/schemas/Version'
         '404':
           description: Version not available.
+  /health:
```
forus: As it's spring boot functionality that we document here I recommend mention that in the description and add a link to the official documentation.

transmart-api-server/grails-app/conf/application.yml
```diff
@@ -23,6 +23,14 @@ endpoints:
     enabled: false
     jmx:
         enabled: true
+    health:
```
forus: It looks like `management.security.enabled` and `endpoints.health.sensitive` detect what behavior to be for unauthenticated or authenticated users https://docs.spring.io/spring-boot/docs/1.5.3.RELEASE/reference/html/production-ready-monitoring.html#production-ready-health-access-restrictions
I assume that now we have:
`management.security.enabled=true`
`endpoints.health.sensitive=false`
Maybe it makes sense to make it explicit here what state of flags are and why it's important.

open-api/swagger.yaml
ewelinagr: good idea!

transmart-api-server/grails-app/conf/application.yml
ewelinagr: there is some issue with `endpoints.health.sensitive`, changing it does not affect the behavior at all, so I would rather not add it here. But regarding `endpoints.health.sensitive` you're right. Apparently the defaults differ between spring-boot-actuator version, at least according to the documentation. I'll add it here.

PR #451, opened 2019-08-14
Deny users without roles

gijskant:


transmart-api-server/src/main/groovy/org/transmartproject/api/server/server/TransmartKeycloakAuthenticationProcessingFilter.groovy
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * This filter denies access to users without any roles assigned to them.
+ */
+@Slf4j
+class TransmartKeycloakAuthenticationProcessingFilter extends KeycloakAuthenticationProcessingFilter {
+
+    TransmartKeycloakAuthenticationProcessingFilter(AuthenticationManager authenticationManager) {
+        super(authenticationManager)
+        log.info("Enable filter that denies access to users without any roles assigned to them.")
```
ewelinagr: other token-verification logs (e.g. from the keycloak-client plugin) are on the error level, maybe useful to have this one as error too

transmart-api-server/src/main/groovy/org/transmartproject/api/server/server/TransmartKeycloakAuthenticationProcessingFilter.groovy
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Override
+    Authentication attemptAuthentication(HttpServletRequest request, HttpServletResponse response) throws AuthenticationException, IOException, ServletException {
+        Authentication authentication = super.attemptAuthentication(request, response)
+        if (authentication && authentication.authorities.empty) {
```
ewelinagr: works fine, but grants access to any role, also an invalid one (other than ROLE_ADMIN, ROLE_PUBLIC and transmart-format specific)

transmart-api-server/src/main/groovy/org/transmartproject/api/server/server/TransmartKeycloakAuthenticationProcessingFilter.groovy
gijskant: But error level should only be used for errors, right? these are not errors.

transmart-api-server/src/main/groovy/org/transmartproject/api/server/server/TransmartKeycloakAuthenticationProcessingFilter.groovy
gijskant: The authentication validation now parses the authorities to TranSMART access levels.

Issue #434, opened 2019-07-08
Update concept name instead of concept path


gijskant:
Updating concept path at upload (as happens here: https://github.com/thehyve/transmart-core/blob/dev/transmart-copy/src/main/groovy/org/transmartproject/copy/table/Concepts.groovy#L88) seems dangerous, as it is the primary key. Updating the concept name, however, would be very useful.

gijskant:
Fixed by #469.

PR #469, opened 2019-10-14
Detect concept name updates when loading data to concept dimension table

ewelinagr:
Fixes: [TMT-1028](https://jira.thehyve.nl/browse/TMT-1028)


transmart-copy/src/main/groovy/org/transmartproject/copy/table/Concepts.groovy
```diff
@@ -101,26 +106,49 @@ class Concepts {
         conceptCodeToConceptPath[conceptCode] = conceptPath
     }
 
+    void updateNameForConcept(
```
gijskant: Can this function be merged with `updatePathForConcept`?
The delete line can be simplified to `"delete from ${TABLE} where concept_cd = :conceptCode"`.

gijskant:
Nice that the test first tests that it fails without `-U`, and then checks the successful update with the parameter.

gijskant:
Fixes #434.

PR #473, opened 2019-10-30
Made BIOMART_PASSWORD configurable

### LINKED ISSUE #472: Db user credentials not handled correctly in Docker enrypoint
There are 2 variables in transmart-api-server entrypoint for tranSMART db user, that are not used:
https://github.com/thehyve/transmart-core/blob/dev/docker/transmart-api-server/docker-entrypoint.sh#L27
As a result, the default values set [here](https://github.com/thehyve/transmart-core/blob/dev/transmart-api-server/grails-app/conf/application.yml#L47) are always used for transmart-api-server.

These values should be either set using environment variables and then added to dataSource properties in the enrtypoint file, or they should be removed.

TG908:
Fixes #472

I still need to test this! Will update the PR once I got the chance to test this.

### DIFF
--- a/docker/transmart-api-server/docker-entrypoint.sh
+++ b/docker/transmart-api-server/docker-entrypoint.sh
@@ -35,7 +35,7 @@ dataSource:
     driverClassName: org.postgresql.Driver
     dialect: org.hibernate.dialect.PostgreSQLDialect
     url: jdbc:postgresql://${PGHOST}:${PGPORT:-5432}/${PGDATABASE:-transmart}?currentSchema=public
-    username: ${BIOMART_USER:-biomart_user}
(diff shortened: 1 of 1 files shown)


TG908:
Testing is done. It works.

TG908:
any updates on this?

gijskant:
Thanks for the change. One remark: could you please name the variable `BIOMART_USER_PASSWORD` or `BIOMART_USER_PWD` for consistency with, e.g., https://github.com/thehyve/transmart-core/blob/dev/transmart-data/vars.sample#L84.
The `biomart` schema/user and the `biomart_user` schema/user are different things.
When changed, we will soon merge your two PRs for making the database password configurable.

PR #475, opened 2019-11-17
Add PATIENT_COUNT_THRESHOLD

TG908:
I added the option to configure the `PATIENT_COUNT_THRESHOLD ` in `docker-entrypoint.sh`

### DIFF
--- a/docker/transmart-api-server/docker-entrypoint.sh
+++ b/docker/transmart-api-server/docker-entrypoint.sh
@@ -47,6 +47,9 @@ org.transmartproject.system.writeLogToDatabase: false
 # If true, deny access to users without any role
 org.transmartproject.security.denyAccessToUsersWithoutRole: ${DENY_ACCESS_WITHOUT_ROLE:-false}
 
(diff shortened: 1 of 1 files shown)


TG908:
any updates on this?

gijskant:
Hi Tim,

We're curious why you need this option to be available? We deliberately left more advanced configuration options out to keep deployments as simple as possible, to avoid users running in complex issues (why can't I see the data I expected) and to keep configuration in Keycloak simple.

This particular feature, patient counts with thresholds, requires administrators to create roles per study and add users to these roles. If users have mixed permissions (full access on one study, limited counts on another), this can lead to confusing results. Also the performance is worse when using these thresholds.
Although the counts with threshold limit what users can see, it does not provide strong privacy guarantees like k-anonymity (see the documentation).

Kind regards,
Gijs Kant

TG908:
Thanks for your reply. I just read the documentation of transmart-api-server and thought this was missing. It makes sense to keep the deployment as simple as possible.

PR #477, opened 2019-12-06
Upgrade Grails to 3.2.12.

gijskant:


ewelinagr:
I noticed there are grails versions defined in plugin descriptors: [TransmartRestApiGrailsPlugin](https://github.com/thehyve/transmart-core/blob/update-dependencies/transmart-rest-api/src/main/groovy/org/transmartproject/rest/TransmartRestApiGrailsPlugin.groovy#L37), [TransmartCoreDbTestsGrailsPlugin](https://github.com/thehyve/transmart-core/blob/update-dependencies/transmart-core-db-tests/src/main/groovy/transmart/core/db/tests/TransmartCoreDbTestsGrailsPlugin.groovy#L25), [TransmartCoreGrailsPlugin](https://github.com/thehyve/transmart-core/blob/update-dependencies/transmart-core-db/src/main/groovy/org/transmartproject/db/TransmartCoreGrailsPlugin.groovy#L37). Probably just some unused information, not sure if it should be updated/removed.

PR #479, opened 2020-02-10
Grails upgrade; publish releases from Travis

gijskant:
- Upgrade Grails version to 3.2.13.
- Release transmart-core to Nexus, Docker Hub from Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,15 @@
+os: linux
 dist: trusty
-sudo: required
-
 language: groovy
-
-jdk:
-  - oraclejdk8
-php:
-  - 5.6
+jdk: oraclejdk8
+php: 5.6
 env:
   global:
     - TERM=dumb
     - PGCONFIG_PARAMS="PG_CONFIG=/usr/lib/postgresql/9.6/bin/pg_config"
-  matrix:
-    - DIR=transmart-core-api; TEST=check
-    - DIR=transmart-core-db-tests; PREPARE=assemble; TEST=check
-    - DIR=transmart-rest-api; TEST=check
-    - DIR=transmart-core-db; INSTALL=":transmart-data:createVars :transmart-data:setupPostgresTest"; TEST=check
-    - DIR=transmart-api-server; TEST=check
-    - DIR=transmart-copy; INSTALL=":prepareTestDatabase"; PREPARE=shadowJar; TEST=check
-    - DIR=transmart-schemas; PREPARE=assemble; TEST=check
-    - DIR=.; TEST="publishToMavenLocal"
 
 addons:
-  postgresql: 9.6
+  postgresql: '9.6'
 
 before_cache:
   - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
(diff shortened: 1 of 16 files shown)


docker/transmart-api-server/Dockerfile
```diff
@@ -1,18 +1,18 @@
...
 
-# The port is configured in the docker-entrypoint.sh when writing the config.yml
-EXPOSE 8081
+ARG TRANSMART_VERSION="17.2.6"
 
-ENV TRANSMART_VERSION "17.2.6"
 ENV TRANSMART_USER_NAME transmart
 ENV TRANSMART_GROUP_NAME "${TRANSMART_USER_NAME}"
 ENV TRANSMART_USER_HOME "/home/${TRANSMART_USER_NAME}"
 ENV TRANSMART_API_SERVER_WAR_URL  "https://repo.thehyve.nl/service/local/repositories/releases/content/org/transmartproject/transmart-api-server/${TRANSMART_VERSION}/transmart-api-server-${TRANSMART_VERSION}.war"
```
ewelinagr: This one is not used anymore, so it could be removed

transmart-api-server/src/main/groovy/org/transmartproject/api/server/user/AccessLevels.groovy
```diff
@@ -84,7 +84,7 @@ class AccessLevels {
             return true
         }
         Map<String, PatientDataAccessLevel> studyToPatientDataAccessLevel = buildStudyToPatientDataAccessLevel(authorities)
-        return !studyToPatientDataAccessLevel.empty
```
ewelinagr: is this not working only on Map objects? There are other places where this is used, but the cases I found are on List, Set and String objects

transmart-schemas/src/main/resources/db/changelog/changes/20200210135535_autoincrement_id_columns.xml
```diff
@@ -0,0 +1,43 @@
...
+        <addAutoIncrement schemaName="i2b2demodata" tableName="patient_dimension" columnName="patient_num" columnDataType="integer"/>
+
+        <comment>Add autoincrement to relation type identifier column.</comment>
+        <addAutoIncrement schemaName="i2b2demodata" tableName="relation_type" columnName="id" columnDataType="integer"/>
+
+        <comment>Add autoincrement to the metadata tag identifier columns.</comment>
+        <addAutoIncrement schemaName="i2b2metadata" tableName="i2b2_tag_types" columnName="tag_type_id" columnDataType="integer"/>
+        <addAutoIncrement schemaName="i2b2metadata" tableName="i2b2_tag_options" columnName="tag_option_id" columnDataType="integer"/>
+        <addAutoIncrement schemaName="i2b2metadata" tableName="i2b2_tags" columnName="tag_id" columnDataType="integer"/>
+    </changeSet>
```
ewelinagr: what about `i2b2demodata.async_job.id`?

gijskant: That one is handled differently. The domain class in the Grails application is configured to use a sequence explicitly to generate identifiers: https://github.com/thehyve/transmart-core/blob/dev/transmart-core-db/grails-app/domain/org/transmartproject/db/job/AsyncJobCoreDb.groovy#L41.
The other tables are populated by transmart-copy, which expects identifiers that are not explictly set (like concept_cd) to be autogenerated by the database.

transmart-api-server/src/main/groovy/org/transmartproject/api/server/user/AccessLevels.groovy
gijskant: I'm not sure...

transmart-schemas/src/main/resources/db/changelog/changes/20200210135535_autoincrement_id_columns.xml
ewelinagr: ah ok!

transmart-api-server/src/main/groovy/org/transmartproject/api/server/user/AccessLevels.groovy
gijskant:
```
groovy:000> [].empty
===> true
groovy:000> ['a'].empty
===> false
groovy:000> [:].empty
===> null
groovy:000> "".empty
===> true
groovy:000> ([] as Set).empty
===> true
```
Seems to only fail on maps indeed

gijskant: I guess it conflicts with the syntactic sugar of addressing elements: `dict.key === dict[key]`, so `dict.empty` means `dict['empty']`...

.travis.yml
```diff
@@ -40,33 +27,149 @@ before_install:
...
+        - popd
+
+    - name: transmart-rest-api
+      env: DIR=transmart-rest-api
+      script:
+        - pushd "$DIR"
+        - $GRADLE_SCRIPT check
+        - popd
+
+    - name: transmart-core-db on PostgreSQL 9.6
```
ewelinagr: 9.6 or 9.2? Or it is not related with postgresql-9.2 that is removed later?

gijskant: `9.2` is default, but we configured `9.6` to be used instead. Up to version `10`, choosing a database version is relatively easy, for `11` you need to do more effort, such as actually removing `9.2`.

PR #484, opened 2020-06-26
Update to Grails 3.3.11; update dependencies.

gijskant:


build.gradle
```diff
@@ -19,44 +19,38 @@ buildscript {
...
-        classpath 'org.grails.plugins:hibernate5:6.0.4'
+        classpath 'org.grails.plugins:hibernate5:6.1.12'
         classpath "com.google.protobuf:protobuf-java:${protobufVersion}"
         classpath "com.google.protobuf:protobuf-java-util:${protobufVersion}"
-        classpath 'org.grails.plugins:views-gradle:1.0.12'
+        classpath 'org.grails.plugins:views-gradle:1.2.10'
         classpath 'net.linguica.gradle:maven-settings-plugin:0.5'
         classpath "org.fusesource.jansi:jansi:${jansiVersion}"
-        classpath 'org.owasp:dependency-check-gradle:3.2.1'
+        classpath 'org.owasp:dependency-check-gradle:5.3.2.1'
```
ewelinagr, line 28: This doesn't look like a valid version, should it stay `3.2.1`?

transmart-core-db-tests/src/integration-test/groovy/org/transmartproject/db/user/UserAccessLevelSpec.groovy
```diff
@@ -19,8 +19,8 @@
 
 package org.transmartproject.db.user
 
-import grails.test.mixin.integration.Integration
-import grails.transaction.Rollback
+import grails.testing.mixin.integration.Integration
+import grails.gorm.transactions.Rollback
```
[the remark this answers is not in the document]
gijskant: `grails.transaction.Rollback` is deprecated (http://docs.grails.org/3.3.11/api/grails/transaction/Rollback.html), but it still works now. For upgrading to Grails 4, it needs to be updated everywhere.

build.gradle
gijskant, line 28: It is the latest version: https://mvnrepository.com/artifact/org.owasp/dependency-check-gradle

ewelinagr: you're right, but I have some issue with this version, I'm trying to find out why:
```
FAILURE: Build failed with an exception.

* Where:
Build file '/home/ewelina/transmart-core/build.gradle' line: 34

* What went wrong:
A problem occurred evaluating root project 'transmart-core'.
> Failed to apply plugin [id 'org.owasp.dependencycheck']
   > Could not create task of type 'Analyze'.
```

build.gradle
```diff
@@ -19,44 +19,38 @@ buildscript {
...
+    quartzVersion = '2.3.2'
     springSecurityPluginVersion = '3.2.3'
-    springSecurityVersion = '4.1.4.RELEASE'
-    keycloakVersion = '6.0.1'
-}
-
-task wrapper(type: Wrapper) {
-    gradleVersion = gradleWrapperVersion
+    springSecurityVersion = '4.2.17.RELEASE'
+    keycloakVersion = '8.0.2'
```
ewelinagr, line 53: Why this version? Unfortunately there is a bug introduced in v. 8.x - https://issues.redhat.com/browse/KEYCLOAK-10595 and the application doesn't start. It is fixed in version > 9 (I tested with 9.0.3). Could we use a newer than 8?

build.gradle
ewelinagr: This issue is there only when running the application in development mode (using `grails run-app`), it works fine when running as a .jar. I still haven't found a fix for this.

gijskant: For me the plugin fails with gradle 4, but works with gradle 3.5.1. Weird. I never use `grails run-app` actually.

build.gradle
gijskant: I had issues with tests with version `10`, nice that version 9 works. I updated the dependency.

build.gradle
gijskant: Its does work for me with `grails run-app` and `gradle bootRun` with Grails 3.3.11, Gradle 4.10.3 and Java 8. Could you check the versions?

ewelinagr: I have the same versions for all and the gradle command works for me as well. It's only `grails run-app` that fails, but if it works for you, it should be fine, maybe it's just my local issue then.

transmart-core-db/src/main/groovy/org/transmartproject/db/util/HibernateUtils.groovy
```diff
@@ -105,6 +106,7 @@ class HibernateUtils {
         toFix.each {
             instanceFetchers[it.substring(3)] = instanceFetchers.remove(it)
         }
+        */
```
ewelinagr, line 109: Could this workaround be removed?

gijskant: Yes, it seems to work fine without.

